PR #60, opened 2022-03-17
Jupyter Notebook having issues with data directory workaround

cvoegele:
Adds a function to fix the data directory not being used inside of the jupyter notebook.
There is also documentation on this in the container.md.
Also updates the container builds to reflect the general installation process of karabo now.


Lukas113:
This looks fine I think. Alternatively we could have just download the data as described in https://ska-telescope.gitlab.io/external/rascil/RASCIL_install.html. But that's probably not necessary for our use-case.

PR #61, opened 2022-03-17
Fix Installation Issues with RASCIL and BDSF

cvoegele:
Generally the issues have been fixed with updates to the repositories i4ds/bdsf-build and i4ds/rascil-build. 
The bdsf build has been updates so that the import error on importing `from rascil.apps import rascil_imager` no longer fails.
The issue lied in the aotools package, which downgraded upon installation the numpy installation, which then had issues, as the bdsf was installed against a certain numpy instance with boost-numpy. 
Now the bdsf package has different versions for different numpy versions. 

The rascil imager is now running again on python 3.8 (generally python 3.8 is the version we use now)
@Lukas113 please test this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ This pipeline serves as the starting point for the SKA Digital Twin Pipeline, wh
 ## Local Installation
 #### Conda
 ```shell
+conda create -n karabo python=3.8
+conda activate karabo
 conda install -c i4ds -c conda-forge karabo-pipeline=0.2.0
 ```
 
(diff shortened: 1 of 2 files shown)


Lukas113:
I will test it. I had issues with this using python 3.9, but with 3.7 it worked.

Lukas113:
`from rascil.apps import rascil_imager` didn't cause an error caused by BDSF using python 3.8 and the installation procedure of this branch.

PR #73, opened 2022-03-31
fix doc creation

cvoegele:
The action now deletes previous folders so that the error of folders already existing goes away.


Prazn:
[Migrate dem.service.ts](https://trello.com/c/x3t87EKw/1692-migrate-demservicets)

PR #152, opened 2022-09-23
Modified emv yaml with correct xarray version.

kenfus:
- Fixed xarray to 2022.3.0


Lukas113:
I don't want to accept this pull request since xarray is a dependency of RASCIL (see https://github.com/i4Ds/rascil-build/blob/main/conda/meta.yaml) and it should be made there. Therefore I suggest to reject this request.

PR #185, opened 2022-10-27
Writing API for the long observations!

rohitcbscient:
The workflow is re-distributed for the long hours of observations into suitable modules.


Lukas113 (changes requested):
`create_vis_long`
- replace magic numbers and objects and e.g. pass them as arguments if a fast fix is needed. Then, just create the Objects outside with the parameters that are fixed and add the day dependent parameters through a `set` method. But I think for the future and for consistency we should create a new class like `ObservationLong` and handle everything there.
- `BeamPattern` is not defined
- replace magic path './karabo/test/data/beam_vis_' prefix with path from `karabo.util.FileHandle.py`  `FileHandle` and let the Handler take care of creating and removing the dirs. If needed, enhance `FileHandle`
- replace `save_meerkat_cst_file` with any telescope. maybe as callback of BeamPattern through `input_telescope`?

The rest seems fine!

Lukas113 (approved):

PR #186, opened 2022-10-27
Adding functionality for baseline length based in the telescope module.

rohitcbscient:
Addressing the baseline-based filtration for the stations of the telescope. Both lower and upper cut-off on the baseline is implementation.

### DIFF
--- a/karabo/simulation/telescope.py
+++ b/karabo/simulation/telescope.py
@@ -6,7 +6,7 @@ from typing import List
 
 import numpy as np
 import oskar.telescope as os_telescope
-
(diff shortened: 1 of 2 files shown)


Lukas113 (approved):
The code seems to be good. Unfortunately, there are no typings in the definition of the added methods. But since we are not consistent in this respect throughout the framework, I do not request any changes.

deiruch:
@Lukas113 100% agree with your comment. What do you think if only merge new code *WITH* proper type annotations going forward, so that things at least do not get worse?

PR #202, opened 2022-11-18
Vt 112 add interface for parallel computing

kenfus:
- Improved usage of the dask client and logic behind it
- Added GPU support for rascil via the ska-cuda-nifty-gridder package
- Created conda package for ska-cuda-nifty-gridder package
- Created new tests and refactored the tests a bit.

ska-cuda-nifty-gridder package is created [here](https://github.com/i4Ds/ska-gridder-nifty-cuda_build), but it's not as elegant yet as other packages created by @Lukas113. The question is also if we need this because I don't think it's as actively developed as rascil or oskar:

https://github.com/i4Ds/ska-gridder-nifty-cuda_build

Also, for some reason, the GitHub action [Build Dev Env](https://github.com/i4Ds/Karabo-Pipeline/actions/runs/3498022327) is unable to echo environment variables defined in `environment-dev.yaml`.


Lukas113 (approved):
looks fine

PR #222, opened 2022-12-09
Updated logic on GPU usage if use_gpu is set to None.

kenfus:
- Changed test to automatically decide if it should use CPU or GPU
- Updated logic by only deciding what to use when use_gpu is set to None.


kenfus:
@Lukas113 please review and approve.

Lukas113 (approved):

Issue #232, opened 2022-12-13
astropy 5.2 ModuleNotFoundError


Lukas113:
I've setup a new environment using `environment-dev.yaml` and encountered a `ModuleNotFoundError: No module named 'astropy._erfa'` Error when `import rascil` (`rascil==0.7.0) was called. This was caused by `astropy` where the version was `astropy==5.2` (release 12.12.2022). However, as soon as I've downgraded to `astropy==5.1.1` the error didn't occur anymore.

Since the error is thrown through RASCIL I think we should specify the version also in the RASCIL build (and if possible not in  `environment-dev.yaml`). In addition, we have to check whether it is also an issue in RASCIL 0.7.1 #157 )

Lukas113:
![image](https://user-images.githubusercontent.com/32157249/207357224-526a2c54-9dac-4a3a-91fb-da502a567f93.png)

kenfus:
This is now updated via Rascil and #236 will include the removal of astropy as dependency in Karabo. The version of astropy will be defined only in our rascil-build.

Lukas113:
Nice. I merged the PR and therefore close this issue.

PR #236, opened 2022-12-16
Oskarpy 2.8.3 works again

kenfus:
- to give a cuda-toolkit version, you cannot do cuda-toolkit=11.7.0 but you need to do it as follows: `conda install -c "nvidia/label/cuda-11.7.0" cuda-toolkit`
- Fixed astropy to a version because 5.2 seems broken.
- Rebuilt oskar without numpy fixing because this caused errors when updating oskar to 2.8.3.

### DIFF
--- a/.github/workflows/conda-build.yml
+++ b/.github/workflows/conda-build.yml
@@ -16,6 +16,7 @@ jobs:
         run: |
           conda config --append channels conda-forge
           conda config --append channels i4ds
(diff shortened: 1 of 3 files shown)


Lukas113:
@kenfus I think it also makes sense to fix it in the karabo build `meta.yaml`!? This PR closes #232

Lukas113 (dismissed):

Lukas113 (approved):

PR #245, opened 2022-12-21
Js return indices of filter functions

jejestern:
These changes make it possible that we can also get back the indices from the filtered sky in fiter_by_radius().


Lukas113 (changes requested):
Changes are looking almost fine. We decided to just approve further PR with type annotations. So please replace `indices=False` with `indices: bool = False,`. Afterwards I'll merge it.

jejestern:
Thank you for the information. I now added the type annotation.

Lukas113 (approved):

PR #277, opened 2023-01-20
Lg 210 detect sources in image

### LINKED ISSUE #210: `SourceDetectionResult.detect_sources_in_image()` does not fail when no sources are detected.
```
Also note that I found out that Karabo wasn’t handling correctly the case where it can’t detect any
source. Indeed when Karabo perform the detection, It then extract the pixel coordinates using
indexes (If I did understand the source code right). But this extraction expects to obtain an array that
is 2-dimesionnal (one list of ‘x’ coordinates and one of ‘y’ coordinates). But when there is no sources
detected, we obtain simply an empty array (represented like so ‘[ ]’) and so it has only 1 dimension.
This causes an ‘IndexError’ error when running the function
‘SourceDetectionResult.detect_sources_in_image()’ and that no source is detected.
```

- Patrick Arlt

Lukas113:
- Removed `detect_sources_in_dirty_image` because too much redundancy with `detect_sources_in_image`, thus is no longer dependent on `Imager`.
- No lazy-loading of header and data from `Image`. So an instance variable is always accessed directly. That means you always know where the information comes from.
- An `Image` is now always loaded from a file when instantiation is performed. That means, the image instance is not only used to get the file-path and save something there (like before) but the file has to be saved first, so that an `Image` instance can be created.
- `SourceDetectionResult.detect_sources_in_image` catches `bdsf RuntimeError` if all pixels in the image are blanked.
- Bugfix when just a single source is present for `SourceDetectionResult.detect_sources_in_image` and `sky.explore_sky`.
- Relocated `image_header_has_parameters` from `data_utils` since this function belongs clearly to an `Image`.
- made `__transform_bdsf_to_reduced_result_array` robust to unexpected inputs, fixes #210
- Additional minor adjustments line fixing typings
- Adapted according interfaces and tests


karabo/test/test_source_detection.py
```diff
@@ -61,6 +65,38 @@ def test_source_detection_plot(self):
...
+        imaging_npixel = 2048
+        imaging_cellsize = 3.878509448876288e-05
+        imager_askap = Imager(
+            visibility_askap,
+            imaging_npixel = imaging_npixel,
+            imaging_cellsize = imaging_cellsize,
+        )
+        image_blanked = imager_askap.get_dirty_image()
+        beam_guess = (0.06414627663254034, 0.05891435806172773, 69.63573045562626)
+        PyBDSFSourceDetectionResult.detect_sources_in_image(image=image_blanked, beam=beam_guess)
```
kenfus, line 99: Maybe check the output against something? E.g. sum(sources) == 0

kenfus (changes requested):

karabo/test/test_source_detection.py
Lukas113: I've adapted the test partially. Partially only, because if the image pixels are blanked, the return value is not a `bdsf_image` but it throws a `RuntimeError` (which is caught in `detect_sources_in_image` and therefore it then just returns None). Therefore I just want to check whether the specific error is caught successfully. To check this you only have to call `PyBDSFSourceDetectionResult.detect_sources_in_image` with a blanked image and nothing more. In addition to the request I've implemented if the return value is None.

kenfus (approved):

PR #278, opened 2023-01-20
Vt 261 imager overwrites cellsize

kenfus:
- added grids to the plot to show that the coordinates are rounded and not only vertical or horizontal
- steamlined parameters
- RASCIL now accepts booleans and not only strings (e.g. 'True' vs True in python).
- Added tests for the cellsize parameter.


karabo/imaging/image.py
```diff
@@ -71,17 +71,17 @@ def get_squeezed_data(self) -> NDArray[np.float64]:
 
     def plot(
         self,
-        title: str,
+        title: Optional[str] = None,
         xlim: Optional[Tuple[float, float]] = None,
         ylim: Optional[Tuple[float, float]] = None,
         figsize: Optional[Tuple[float, float]] = None,
-        plot_title: Optional[str] = None,
```
rohitcbscient, line 78: okokokok

rohitcbscient (approved):

Issue #296, opened 2023-02-03
`Running an interfoerometer simulation` from documentation should run through


deiruch:
Executing the example on CSCS after following the steps in https://github.com/i4Ds/Karabo-Pipeline/issues/283#issuecomment-1416242369 results in the following:
```
>>> from karabo.simulation.sky_model import SkyModel
The RASCIL data directory is not available - continuing but any simulations will fail
>>> from karabo.simulation.telescope import get_OSKAR_Example_Telescope
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ImportError: cannot import name 'get_OSKAR_Example_Telescope' from 'karabo.simulation.telescope' (/scratch/snx3000/sfelix/miniconda3/envs/karabo-env/lib/python3.9/site-packages/karabo/simulation/telescope.py)
>>> from karabo.simulation.interferometer import InterferometerSimulation
>>> from karabo.simulation.observation import Observation
>>> import numpy as np
>>> 
>>> # create a simple sky model with three point sources
>>> sky = SkyModel()
>>> sky_data = np.array([
...             [20.0, -30.0, 1, 0, 0, 0, 100.0e6, -0.7, 0.0, 0, 0, 0],
...             [20.0, -30.5, 3, 2, 2, 0, 100.0e6, -0.7, 0.0, 600, 50, 45],
...             [20.5, -30.5, 3, 0, 0, 2, 100.0e6, -0.7, 0.0, 700, 10, -10]])
>>> sky.add_points_sources(sky_data)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'SkyModel' object has no attribute 'add_points_sources'
>>> 
>>> # get different predefined telescopes, like here the OSKAR example telescope.png, with a handy functions
>>> telescope = get_OSKAR_Example_Telescope()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'get_OSKAR_Example_Telescope' is not defined
>>> # overwrite or set any of the implemented configuration values
>>> telescope.centre_longitude = 3
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'telescope' is not defined
>>> 
>>> simulation = InterferometerSimulation("./test_result.ms")
>>> 
>>> # create new observational settings with most settings set to default except the start frequency set to 1e6
>>> observation = Observation(1e6)
use_gpu is None, using is_cuda_available() to set use_gpu to True
[...]

PR #299, opened 2023-02-06
Vt 296 running an interfoerometer simulation from documentation should run through

kenfus:


Lukas113 (approved):
I haven't looked at the details of the implementation but everything according to issue #296 is addressed. I also like the new documentation structure defined in `index.rst`.

PR #324, opened 2023-02-16
Rs 290 no error when combining visibilities after using observationlong

rohitcbscient:
Bugfixes


karabo/simulation/visibility.py
```diff
@@ -200,26 +213,111 @@ def combine_vis(
...
+                        block.num_channels,
+                        block.num_baselines,
+                        out_vis[j][t],
+                    )
+        test = 0
+        if test:
+            num_times = out_vis[j].shape[0]
+            for t in range(num_times):
+                # Dummy data to write.
+                time_stamp = time_start[0] + t * time_inc[0] / 86400.0
```
Lukas113, line 255: To me this doesn't look useful because `if test:` is always False. In addition, if this is a real test, I think it shouldn't be part of the source-code.

Lukas113 (changes requested):
The changes are looking good except of the objection in `visibility.py 250`.

What could be improved is using type annotations for return types (e.g. `combine_vis) and for data-structures like `visibility.combine_vis out_vis` (where you first create a list of integers and later you have a list of arrays which is something you shouldn't actually do). But because this is something that was not changed in this PR I don't request changes for this.

karabo/simulation/visibility.py
kenfus: @rohitcbscient Any updates on this? As soon as this is merged, I can start timing how quick Oskar is on the GPU! I can support, if you would like

rohitcbscient:
The test is updated.

Lukas113 (approved):

Issue #313, opened 2023-02-10
Dask version number fixed


deiruch:
Depending on the internal state of the conda solver, we either get Dask 2021.9.1 or 2022.11.0. We should fix it to one or the other, to increase consistency and reliability.

Lukas113:
Done with PR #348

PR #348, opened 2023-03-10
Lg 298 docker

### LINKED ISSUE #298: Docker image contains prepared Jupyter notebook
It would be nice to have in the docker image a prepared notebook with the instructions given in the documentation. "untitled.ipynb" is not facilitating the starting and even confusing as it gives the impression that I am not in the correct path

Lukas113:
**New Docker setup:**
- As discussed in the meeting, the docker image is created for every test, so no prebuild image for testing anymore. Build time is about 10-11 minutes. The according Dockerfile is `docker/Dockerfile.dev`.
- We no longer have `cli`, `jupyter` and `development` images under the same image-name with the according tags. The dev- and user images have now distinct image-names, [karabo-pipeline-dev](https://github.com/i4ds/Karabo-Pipeline/pkgs/container/karabo-pipeline-dev) and [karabo-pipeline](https://github.com/i4ds/Karabo-Pipeline/pkgs/container/karabo-pipeline) for the user images. Do not remove these images until the next Karabo release.
- Dev-images-setup: Usually, the dev-images should get purged after usage, but in case of a manual interrupt or some strange runner-error, it might remain and needs to be purged by hand. However, never remove the last image, since the purging procedure of the following run then fails. The according Dockerfile is `docker/Dockerfile.dev` and installs from the repository files.
- User-images-setup: It will be build right after the `Conda Build` action and creates an image using the on anaconda.org just published karabo package. It is then labelled using the `latest` label and the latest karabo-tag. So a user is now able to install legacy docker-images.
- Test-slow workflow is now integrated in Test
- Unification of endings .yaml to yml
- Updated the documentation to changes
- Included sphinx dependencies in `requirements.txt` to get rid of an additional dev dependency file

I've tested the workflows manually and I hope that there is no issue since testing was somehow cumbersome because the workflows must have a connection to the default branch for manual testing.

closes #298


kenfus:
It seems that the solution with the docker image for the testing causes a lot of problems. Why exactly do we need to build a docker image for testing, why not install karabo directly on the GitHub runner and run the tests?

deiruch:
> It seems that the solution with the docker image for the testing causes a lot of problems. Why exactly do we need to build a docker image for testing, why not install karabo directly on the GitHub runner and run the tests?

1. We don't need Docker for every test, we can use Karabo directly.
2. We'll want to test the release images that we produce once a month.
3. To catch problems early, it'd be nice to test images a bit before - to make sure we don't release broken images.

It follows, that
a) it's enough to test released images, once a month
b) it'd be nice to test more often

deiruch:
I wonder whether we need the "latest user image". Users should install released images, right?

Lukas113:
> I wonder whether we need the "latest user image". Users should install released images, right?

As discussed last Friday, the images can have multiple tags and having a latest tag with a release-tag is how most docker-images are released.

Lukas113:
> It seems that the solution with the docker image for the testing causes a lot of problems. Why exactly do we need to build a docker image for testing, why not install karabo directly on the GitHub runner and run the tests?

It can have some problems, but they shouldn't occur very frequently. I just know of 1 case as far where the ghcr.io registration failed. I like the setup with the docker because now the environment can be used for different workflows by just following the setup template of `Build_Dev_Image` and `Purge_Dev_Image`. In addition, we could also use the docker-image for singularity containers for testing in the future. So to conclude, it is more a thought of how this can be used in the future rather than it is a necessity.

Lukas113:
> > It seems that the solution with the docker image for the testing causes a lot of problems. Why exactly do we need to build a docker image for testing, why not install karabo directly on the GitHub runner and run the tests?
> 
> 1. We don't need Docker for every test, we can use Karabo directly.
> 2. We'll want to test the release images that we produce once a month.
> 3. To catch problems early, it'd be nice to test images a bit before - to make sure we don't release broken images.
> 
> It follows, that a) it's enough to test released images, once a month b) it'd be nice to test more often

Regarding to your 3. point: I don't think that it is helpful to make a pre-release, because there could be changes which would not be in the released Docker image. However, what I can make is to run the unit-tests during the docker-build process of the user-image.

Lukas113:
After some sweat and tears, the user setup is now as follows:

The `Build User Image` workflow passes the last tag of the Karabo repository to the `Build and push Docker image` step. The `Dockerfile.user` works with this tag to clone the correct version of the Karabo repository as well as install the correct version of the Conda wheel. The image is then only created if `pytest` was successfully created. If not, the workflow can still be started manually afterwards.

The pytest didn't succeed yet. But as far as I've seen this is mainly because of the `ImportError: libfftw3f_threads.so.3: cannot open shared object file: No such file or directory` error.

Lukas113:
Setup is now adapted to what we have discussed at the last meeting. There are also some minor changes in other files, however the major setup change is:
- Tests on the GitHub runners don't use Docker anymore.
- Tests are not done on each push anymore, instead just on a PR or changes on an active PR.
- The jupyter notebooks are not tested on the GitHub runners, because there the error: `Error: The operation was canceled.` occurs with no further information during the notebook tests. That means, the notebook should be tested on the local machine once the changes on your branch are done. However, this should be fixed in the future.
- Even on the local machine, not all notebooks are tested. I excluded the `HIIM_Img_Recovery.ipynb` notebook because it took hours to run.

kenfus (dismissed):

conda/meta.yaml
```diff
@@ -19,20 +19,21 @@ requirements:
...
-    - numpy<=1.22
-    - matplotlib
-    - oskarpy
-    - rascil
+    - oskarpy=2.8.3
+    - rascil=1.0.0
+    - numpy=1.22
     - astropy
-    - requests
+    - dask
```
deiruch, line 26: We should fix the version number, see https://github.com/i4Ds/Karabo-Pipeline/issues/313

Lukas113:
- I've adapted the docker user build to the new setup with the `rel_v{MAJOR}.{MINOR}` setup convention.
- I excluded the tests of the Notebooks on the GitHub runners (see #369 )
- The user docker-images are not tested because tests fail (see #368 )

kenfus (approved):

Issue #233, opened 2022-12-14
RASCIL 1.0.0


Lukas113:
RASCIL has a major release [RASCIL 1.0.0](https://gitlab.com/ska-telescope/external/rascil-main/-/tags) since yesterday (13.12.2022). I think we should build it as soon as possible and try to integrate it to our pipeline.

Lukas113:
Release `1.0.0` still has the issue mentioned in #232 , therefore I've opened an [Issue](https://gitlab.com/ska-telescope/external/rascil-main/-/issues/1) on their repository. But for us it means that we just have to restrict [astropy](https://github.com/astropy/astropy) < 5.2 for our builds.

Lukas113:
I've seen that [RASCIL](https://gitlab.com/ska-telescope/external/rascil-main) has additional [dependencies](https://gitlab.com/ska-telescope/external/rascil-main/-/blob/master/requirements.in): [ska-sdp-datamodels](https://gitlab.com/ska-telescope/sdp/ska-sdp-datamodels), [ska-sdp-func-python](https://gitlab.com/ska-telescope/sdp/ska-sdp-func-python), [mistune](https://anaconda.org/conda-forge/mistune), [aoflagger](https://anaconda.org/pkgw-forge/aoflagger), [graphviz](https://anaconda.org/anaconda/graphviz) and [nbconvert](https://anaconda.org/anaconda/nbconvert).

The non `ska-sdp-*` dependencies seems to have [conda builds](https://anaconda.org/) which I referenced. Otherwise I referenced the git repository. However, the dependency [radler](https://git.astron.nl/RD/radler) (see #123) still hasn't a conda build yet but is also a dependency we haven't covered yet.

kenfus:
I think we should increase this priority because of #261

fschramka:
Conda package for [ska-sdp-datamodels](https://gitlab.com/ska-telescope/sdp/ska-sdp-datamodels)

https://github.com/i4Ds/Karabo-Feedstock/commit/ae80e7e0dc1f455c7ed5c4a0e84707388843147f

fschramka:
Conda package for ska-sdp-func and ska-sdp-func-python are rdy

deiruch:
This means everything is ready?!

fschramka:
rascil is currently building fine :) 

https://github.com/i4Ds/Karabo-Feedstock/actions

[...]

PR #349, opened 2023-03-10
Changed certain parameters in RASCIL, made function `is_cuda_available` clearer and added a NotImplementedError for RASCIL WAGG.

kenfus:

PR #356, opened 2023-03-14
Fs rascil 1 0 0 integration

Lukas113:
tests should fail because we still have legacy test-setup which is using RASCIL 0.7.0 but here we have RASCIL 1.0.0

### DIFF
--- a/environment-dev.yaml
+++ b/environment-dev.yaml
@@ -1,26 +1,24 @@
 name: karabo_dev_env
 channels:
-  - i4ds
+  - i4ds/label/rel_v14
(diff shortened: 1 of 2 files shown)


Lukas113:
closes #233 

`radler` is still not included and I think it should be at some point. Otherwise we can't use the whole RASCIL functionality.

fschramka:
what do you want to add before we can close this? radler?

kenfus:
Yep, those I have found up to now: https://github.com/i4Ds/Karabo-Pipeline/pull/349/files#diff-7e41e5fee3ff96bd607ca83ec1c82a5663b584d6808920dac493e4622f0802f2

PR #365, opened 2023-03-22
SKA-LOW corrected layout

rohitcbscient:
Correct SKA-low coordianates


rohitcbscient:
SKA low with the station layouts.

Issue #326, opened 2023-02-20
Implement Ionospheric Screen into Karabo


rohitcbscient:
Oskar has an implementation for inserting time-varying ionospheric screen based on ARastmospy (https://github.com/shrieks/ARatmospy)
Oskar Doc: https://ska-telescope.gitlab.io/sim/oskar/python/example_ionosphere.html?highlight=ionospheric%20screen

chrisfinlay:
We would like to include this into the SDC3a simulations. What is required to get this working? Can we edit the OSKAR settings tree directly to include 'telescope/external_tec_screen/input_fits_file'?

deiruch:
Unit test is failing, will try to merge soon.

PR #366, opened 2023-03-23
Ionospheric screen via ARatmospy fits file in the settings tree.

chrisfinlay:
It doesn’t have a unit test as I’m not sure what an appropriate test would be. 

I think that this should really be set in the observation class rather but OSKAR wants it in the telescope settings tree for some reason. It wasn't playing nice when I tried to do this from the Observation class. 

I think Karabo's preferred solution would be to have ARatmospy as a dependency and make a nice interface to it to generate an ionospheric screen so this is an idea for the future. 

Maybe we can merge this for now as we need it for the SKA data challenge and discuss how this should be ideally implemented.


chrisfinlay:
This is a start towards issue #326.

Lukas113:
@chrisfinlay We are merging the PR into the current release.

For future contribution, we would appreciate it if a PR passes the tests. In this case several `flake8` checks failed (partially also our fault). The easiest way to solve this can be found in [karabo-dev-doc Formatting flake8](https://i4ds.github.io/Karabo-Pipeline/development.html). However, since the release is coming soon, I will take care of it.

Thanks for your contribution.

Lukas113 (approved):

chrisfinlay:
@Lukas113 Great, thanks! Will do in future :)

chrisfinlay:
I just checked and turns out I didn't commit again after running black. Oops!

PR #392, opened 2023-04-13
Adds  method for reshaping.

kenfus:
- Adds method for reshaping an image to a specific dimension.

### DIFF
--- a/environment.yaml
+++ b/environment.yaml
@@ -7,6 +7,7 @@ dependencies:
   - oskarpy=2.8.3
   - rascil=1.0.0
   - numpy=1.22
+  - scipy=1.10.1
   - astropy
   - dask=2022.12.1
   - matplotlib
--- a/karabo/imaging/image.py
+++ b/karabo/imaging/image.py
@@ -3,11 +3,12 @@ from __future__ import annotations
 import logging
 import os
 import uuid
-from typing import Any, Dict, List, Optional, Tuple, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
 import matplotlib
 import matplotlib.pyplot as plt
 import numpy as np
+import scipy
 from astropy.io import fits
 from astropy.wcs import WCS
 from numpy.typing import NDArray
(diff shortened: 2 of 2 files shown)


karabo/imaging/image.py
```diff
@@ -76,6 +77,58 @@ def header_has_parameters(
     def get_squeezed_data(self) -> NDArray[np.float64]:
         return np.squeeze(self.data[:1, :1, :, :])
 
+    def reshape(
+        self,
+        shape: Tuple[int, ...],
+        interpolation_f: Callable[
+            [np.ndarray], np.ndarray
+        ] = scipy.interpolate.RegularGridInterpolator,
```
Lukas113: - use typed NDArray
- use Callabale correctly, it doesn't map the call of `interpolation_f`

karabo/imaging/image.py
```diff
@@ -76,6 +77,58 @@ def header_has_parameters(
     def get_squeezed_data(self) -> NDArray[np.float64]:
         return np.squeeze(self.data[:1, :1, :, :])
 
+    def reshape(
```
Lukas113: This is a resampling method, not reshaping

karabo/imaging/image.py
```diff
@@ -3,11 +3,12 @@
 import logging
 import os
 import uuid
-from typing import Any, Dict, List, Optional, Tuple, Union, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union, cast
 
 import matplotlib
 import matplotlib.pyplot as plt
 import numpy as np
+import scipy
```
Lukas113, line 11: would change to `from scipy.interpolate import RegularGridInterpolator` since when you just do `scipy.interpolate.RegularGridInterpolator` in fun args, it shows `any` to me.

karabo/imaging/image.py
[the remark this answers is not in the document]
kenfus: This was a user request. I agree that people could do it themselves, but it's not as easy as expected. I think it's OK like this and @rohitcbscient liked it too!

environment.yaml
```diff
@@ -7,6 +7,7 @@ dependencies:
   - oskarpy=2.8.3
   - rascil=1.0.0
   - numpy=1.22
+  - scipy=1.10.1
```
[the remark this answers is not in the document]
kenfus: I only put the version because it's the version which is installed anyway by another dependency.

PR #397, opened 2023-04-18
Lg 304 codestyle1

Lukas113:
- telescope.py, beam.py, station.py should be mypy --strict
- beam.py is still under development and therefore it was pretty hard to type. It is possible that I made mistakes there, especially because `astropy.units.Quantity` types work with some `numpy` functions (not all of them), but `numpy` claims that it returns an `NDArray` but it actually returns a `astropy.units.Quantity` if that was the input of at least one variable. But I think it's fine because it is still under development.
- Changed `Telescope.read_OSKAR_tm_file` to meet the true specification of an oskar `position.txt` file: "must have one line with two or three numbers". I have checked the regex against all available `*.tm/position.txt` files. It is also checked against `signed-unsigned`, `scientific-normal`, `float-int`, `single-four+` and `non-numeric-chars` numbers.
- Changed `Telescope.write_to_file` from just copying (not considering current state of object) to save an already available function which considers the current state.


Lukas113:
Here is the code which I tested the regex (didn't want to include in unit-tests because the positive matches are already tested while loading the different telescopes):

```
import re, os, glob

lines = [
    "0 50",
    "-0.1 45",
    "-0.65 -12.12 11",
    "0 50a",
    "0 50 a",
    "0",
    "12 11 14 -4",
    "-6.775379999999999825e+01 -2.302339999999999876e+01",
    "a b",
]
for line in lines:
    print("'"+line.strip()+"'")
    match = re.match(
        r"^\s*([-+]?\d+(?:\.\d+)?(?:[eE][-+]?\d+)?)\s+([-+]?\d+(?:\.\d+)?(?:[eE][-+]?\d+)?)\s*([-+]?\d+(?:\.\d+)?(?:[eE][-+]?\d+)?)?\s*$", line.strip()
    )  # one line with two or three numbers
    if match:
        print('matched')
    else:
        print('not matched')

for tm in glob.glob("karabo/data/*.tm"):
    pos_file = tm+"/position.txt"
    if os.path.exists(pos_file):
        lines = open(pos_file).readlines()
        for line in lines:
            print("'"+line.strip()+"'")
            match = re.match(
                r"^\s*([-+]?\d+(?:\.\d+)?(?:[eE][-+]?\d+)?)\s+([-+]?\d+(?:\.\d+)?(?:[eE][-+]?\d+)?)\s*([-+]?\d+(?:\.\d+)?(?:[eE][-+]?\d+)?)?\s*$", line.strip()
            )  # one line with two or three numbers
            if match:
                numbers = [float(num) for num in match.group().split()]
                long = numbers[0]
                lat = numbers[1]
                alt = 0.0
                if len(numbers) == 3:
                    alt = float(numbers[2])
                telescope = Telescope(long, lat, alt)
                print('matched')
                break
            else:
                print('not matched')
    else:
        print(pos_file+" DOES NOT EXIST")
```

kenfus (approved):

PR #399, opened 2023-04-19
394 skymodel is parallised randomly

kenfus:
- Improves dask support by splitting up the function to use dask on slurm clusters into two parts: If the script runs on a worker node, go into while true-loop and wait for the scheduler file. If on the main node, write the scheduler script and start the dask cluster.
- Improved the FileHandler-Class.
- Removed many "if exist, remove"-randomly trough Karabo
- Removed many reinitialization of telescopes, visibilities or skies, which were there because the FileHandler-Class was not optimal
- Added many tests.
- Added documentation for Dask + Karabo.
- Refactored combine visibilities.


karabo/util/FileHandle.py
```diff
@@ -41,47 +46,49 @@ class FileHandle:
...
+            path = os.path.dirname(path)
+            suffix = ""
+
+        # If a directory is provided, use it as the base path
+        if path:
+            base_path = os.path.abspath(path)
         else:
             base_path = os.path.join(os.getcwd(), "karabo_folder")
+            if suffix.lower() == ".ms":
+                base_path = os.path.join(base_path, str(uuid.uuid4()) + ".MS")
```
Lukas113, line 68: Again, sure you want uppercase?

karabo/util/dask.py
```diff
@@ -1,123 +1,227 @@
...
 
-        client = Client(
-            LocalCluster(n_workers=workers, threads_per_worker=threads_per_worker)
-        )
-    return client
+    dask_client: Optional[Client] = None
+    n_workers_scheduler_node = 1
+    threads_per_worker = 1
+    min_ram_per_worker = None
+    use_dask = None
```
Lukas113, line 47: Missing type-annotations on each variable you assigned `None`. E.g. `dask_client: Optional[Client] = None` is done correctly. Do the same for `min_ram_per_worker` and `use_dask`

karabo/util/dask.py
```diff
@@ -1,123 +1,227 @@
...
+                )
+            else:
+                DaskHandler.dask_client = get_local_dask_client(
+                    DaskHandler.min_ram_per_worker, DaskHandler.threads_per_worker
+                )
 
+        atexit.register(dask_cleanup, DaskHandler.dask_client)
+        return DaskHandler.dask_client
+
+    def should_dask_be_used(override: Optional[bool] = None):
```
Lukas113, line 67: return type `bool` is missing

karabo/util/dask.py
```diff
@@ -1,123 +1,227 @@
...
+            return False
 
-def dask_cleanup(client: Client):
-    # Create the stop_workers file
-    with open(STOP_WORKER_FILE, "w") as _:
-        pass
 
-    # Give some time for the workers to exit before closing the client
-    time.sleep(10)
+def get_local_dask_client(min_ram_gb_per_worker, threads_per_worker) -> Client:
```
Lukas113, line 80: types are missing

karabo/util/dask.py
```diff
@@ -1,123 +1,227 @@
...
-    # Remove the scheduler file
-    if os.path.exists(SCHEDULER_ADDRESS):
-        os.remove(SCHEDULER_ADDRESS)
     if client is not None:
         client.close()
+        client.shutdown()
 
 
-def setup_dask_for_slurm(n_workers_scheduler_node: int = 1):
+def prepare_slurm_nodes_for_dask():
```
Lukas113, line 99: return type `None` is missing

karabo/util/dask.py
```diff
@@ -1,123 +1,227 @@
...
+from typing import Optional
 
 import psutil
 from dask.distributed import Client, LocalCluster
 
-SCHEDULER_ADDRESS = "scheduler_address.json"
-STOP_WORKER_FILE = "stop_workers"
+from karabo.warning import KaraboWarning
 
+DASK_INFO_ADDRESS = os.path.join(".karabo_dask", "dask_info.json")
```
Lukas113, line 15: Shouldn't everything we do regarding directories be in the same directory instead of creating multiple folders on the location a user is using Karabo? In my opinion, we should have a root directory `karabo` and there we can have per task/issue several subdirectories, e.g. `karabo/dask/`.

In my opinion this would be a better solution and should be done with the help of `FileHandler`

karabo/util/dask.py
```diff
@@ -1,123 +1,227 @@
...
+    if is_first_node():
+        # Remove old scheduler file
+        if os.path.exists(DASK_INFO_ADDRESS):
+            os.remove(DASK_INFO_ADDRESS)
 
-        else:
-            # Wait some time to make sure the scheduler file is new
+    else:
+        # Wait some time to make sure the scheduler file is new
+        time.sleep(10)
```
Lukas113, line 116: Seems a bit suspicious to me randomly waiting some time instead of checking if some state is reached or not. However, if you are sure that's a proper solution, we can keep that for the moment.

10 sec looks like a really long time to me. Is this tested? If yes, why is it that long?

karabo/util/dask.py
```diff
@@ -1,123 +1,227 @@
...
+                "--nthreads",
+                str(dask_info["n_threads_per_worker"]),
+                "--nworkers",
+                str(dask_info["n_workers_per_node"]),
+            ]
+        )
+
+        while True:
+            # Wait some time
             time.sleep(5)
```
Lukas113, line 140: This doesn't make any sense to me. Just do a sleep without any chance of changing it's state because nothing happens in the loop looks like something went wrong. If I'm mistaken, an explanation why this is necessary would be helpful.

karabo/util/dask.py
```diff
@@ -1,123 +1,227 @@
...
+        return 1
+    # Calculate number of workers per node
+    ram = psutil.virtual_memory().available / 1024 / 1024
+    n_workers_per_node = int(ram / (min_ram_gb_per_worker))
+    if ram < min_ram_gb_per_worker:
+        KaraboWarning(
+            f"Only {ram} GB of RAM available. Requested at least "
+            f"{min_ram_gb_per_worker} GB. Setting number of "
+            f"workers per node to 1."
+        )
```
Lukas113, line 154: You should be consistent in what units of measurement you operate. E.g. you divide by 1024 and below you throw a warning using GB. GB is in 1000 steps where GiB is in 1024 steps. Therefore I think you either should divide by 1000 above or use GiB in the warning. In addition, it should be consistent with what we use in Karabo. I've only seen GB and not GiB until now. So I think to be consistent you should divide by 1e6.

karabo/util/dask.py
```diff
@@ -1,123 +1,227 @@
...
+        # Removing file
+        if os.path.exists(DASK_INFO_ADDRESS):
+            os.remove(DASK_INFO_ADDRESS)
+        return dask_client
+
+    else:
+        raise Exception("This function should only be reached on the first node.")
 
 
 def get_min_max_of_node_id():
```
Lukas113, line 227: missing type annotation

karabo/util/dask.py
```diff
@@ -134,7 +238,10 @@ def get_min_max_of_node_id():
 
 
 def get_lowest_node_id():
```
Lukas113, line 240: missing type annotation

karabo/util/dask.py
```diff
@@ -134,7 +238,10 @@ def get_min_max_of_node_id():
...
-    return get_min_max_of_node_id()[0]
+    if get_number_of_nodes() == 1:
+        return get_node_id()
+    else:
+        return get_min_max_of_node_id()[0]
 
 
 def get_base_string_node_list():
```
Lukas113, line 247: missing type annotation

karabo/util/dask.py
```diff
@@ -146,26 +253,8 @@ def get_lowest_node_name():
 
 
 def get_number_of_nodes():
```
Lukas113, line 255: missing type annotation.

karabo/util/dask.py
```diff
@@ -146,26 +253,8 @@ def get_lowest_node_name():
...
-    return (
-        get_base_string_node_list()
-        + "["
-        + str(min_node + 1)
-        + "-"
-        + str(max_node)
-        + "]"
-    )
+    n_nodes = os.getenv("SLURM_JOB_NUM_NODES")
+    return int(n_nodes)
```
Lukas113, line 257: seems unsafe since `SLURM_JOB_NUM_NODES` could not be set and therefore `int` casting on NoneType is not allowed.

karabo/util/dask.py
```diff
@@ -146,26 +253,8 @@ def get_lowest_node_name():
...
 
 
 def get_node_id():
```
Lukas113, line 260: missing type annotation

karabo/util/dask.py
```diff
@@ -177,9 +266,5 @@ def is_first_node():
     return get_node_id() == get_lowest_node_id()
 
 
-def get_current_time():
-    return time.strftime("%H:%M:%S", time.localtime())
-
-
 def is_on_slurm_cluster():
```
Lukas113, line 269: missing type annotation

pyproject.toml
```diff
@@ -3,4 +3,9 @@ requires = [
     "setuptools>=42",
     "wheel"
 ]
-build-backend = "setuptools.build_meta"
\ No newline at end of file
+build-backend = "setuptools.build_meta"
+[tool.pytest.ini_options]
+filterwarnings = [
+    "ignore::UserWarning",
+    "ignore::DeprecationWarning",
```
Lukas113, line 10: Sure you want to ignore `DeprecationWarning`? I personally appreciate them since they warn me what could get changed in the future.

Lukas113 (changes requested):

pyproject.toml
kenfus: True, i readded them. I was annoyed by the 123123123123123 of warnings we are getting when running Tests locally. We really need to fix them.

karabo/imaging/image.py
```diff
@@ -68,6 +68,8 @@ def write_to_file(
     ) -> None:
         """Write an `Image` to `path`  as .fits"""
         check_ending(path=path, ending=".fits")
+        if not os.path.exists(os.path.dirname(path)):
+            os.makedirs(os.path.dirname(path))
```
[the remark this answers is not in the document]
kenfus: I think so but I am unsure why i did it like this. I had generally some problem with functions similar to `makedirs`.

karabo/simulation/interferometer.py
```diff
@@ -294,30 +337,32 @@ def __setup_run_simulation_oskar(
                 # Add the remaining sources
                 if len(array_sky) > 0:
                     split_array_sky.append(array_sky)
+                # Delete the array
+                del array_sky
```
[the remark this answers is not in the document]
kenfus: No, apparently not! I guess selecting from a NumPy array does create more than a reference.

karabo/util/dask.py
kenfus: I mean, it only runs once at the beginning, to be 100% sure that it's not reading in an old file. In the next version we will have a better managment for this.

karabo/util/dask.py
kenfus: Yes, absoltely. We need to change the pathing anyway for the next version. I will implement this and fix also some other things

karabo/util/dask.py
kenfus: It's because CSCS intercepts every single python-exit there is and if that is reached on any script on any node, the job ends. Which I guess makes sense? Maybe there is a better solution. However, when the main script reaches a system exit, all the other nodes are also shut-down. I am sure there is a better way, but I did not find it yet.

Lukas113 (approved):

PR #418, opened 2023-05-24
Ionospheric Screen from ARatmospy Implementation

rohitcbscient:
More keywords were added along for the ionospheric screen and simulated it with  ARatmospy.

### DIFF
--- a/karabo/test/test_ionosphere.py
+++ b/karabo/test/test_ionosphere.py
@@ -1,0 +1,136 @@
+import os
+import unittest
+from datetime import datetime, timedelta
+
+import numpy as np
+from ARatmospy.ArScreens import ArScreens
+from astropy.io import fits
+from astropy.wcs import WCS
+
+from karabo.imaging.imager import Imager
+from karabo.simulation.interferometer import InterferometerSimulation
+from karabo.simulation.observation import Observation
+from karabo.simulation.sky_model import SkyModel
+from karabo.simulation.telescope import Telescope
+
+
+class TestSystemNoise(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls) -> None:
+        # make dir for result files
+        if not os.path.exists("result/system_noise"):
+            os.makedirs("result/system_noise")
+
+    def sim_ion(
+        self,
+        screen_width_metres,
+        r0,
+        bmax,
+        sampling,
+        speed,
+        rate,
+        alpha_mag,
+        num_times,
+        frequency,
+        fits_filename,
+    ):
+        """
+        Example from OSKAR
+        """
+        m = int(bmax / sampling)  # Pixels per sub-aperture (200).
+        n = int(screen_width_metres / bmax)  # Sub-apertures across the screen (10).
+        num_pix = n * m
+        pscale = screen_width_metres / (n * m)  # Pixel scale (100 m/pixel).
+        print("Number of pixels %d, pixel size %.3f m" % (num_pix, pscale))
+        print("Field of view %.1f (m)" % (num_pix * pscale))
+        # Parameters for each layer.
+        # (scale size [m], speed [m/s], direction [deg], layer height [m]).
+        layer_params = np.array(
+            [(r0, speed, 60.0, 300e3), (r0, speed / 2.0, -30.0, 310e3)]
+        )
+        my_screens = ArScreens(n, m, pscale, rate, layer_params, alpha_mag)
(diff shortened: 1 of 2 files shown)


Lukas113:
I can't merge this PR. The reason is because the tests fail. The runners are canceled during the tests(see [github-action](https://github.com/i4Ds/Karabo-Pipeline/actions/runs/5209767556/jobs/9421356984)). It is most likely due to memory issues (see [runner operation was canceled](https://github.com/actions/runner/issues/2468) issue). On my local machine, all tests passed, however, not on the GitHub runners. Sadly I wasn't able to estimate the memory consumption of the failing test (also because we have unit-tests instead of pytests, otherwise I could have used [pytest-monitor](https://pypi.org/project/pytest-monitor/)). 

Because the tests succeed on the main-branch and not on this, it has to be the `test_ionosphere` which is the only new test introduced by this PR. However, I don't have a clue how to adapt the test or it's parameters to that the test is still valid and meaningful. Therefore, I suggest that @rohitcbscient should try to adapt the tests so that they work on the GitHub runners. The runners specifications can be seen [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners).

karabo/test/test_ionosphere.py
```diff
@@ -0,0 +1,138 @@
...
+            length=timedelta(hours=0, minutes=10, seconds=1, milliseconds=0),
+            phase_centre_dec_deg=-30.5,
+            number_of_time_steps=5,
+            start_frequency_hz=frequency,
+            frequency_increment_hz=1e6,
+            number_of_channels=1,
+        )
+
+        visibility = simulation.run_simulation(telescope, sky, observation)
+        visibility.write_to_file("./result/test_ion.ms")
```
deiruch, line 131: Is this required? Do we have to write files to disk in order to make it work?

karabo/test/test_ionosphere.py
```diff
@@ -0,0 +1,138 @@
...
+
+        imager = Imager(
+            visibility, imaging_npixel=2048 * 1, imaging_cellsize=5.0e-5
```
deiruch, line 134: To reduce memory size, the image size could be reduced.

karabo/test/test_ionosphere.py
```diff
@@ -0,0 +1,138 @@
...
+            filename=fits_filename, data=data, header=w.to_header(), overwrite=True
+        )
+        return my_screens
+
+    def test_ionosphere(self):
+        # ----------- Ionopsheric Simulations
+        screen_width_metres = 200e3
+        r0 = 5e3  # Scale size (5 km).
+        bmax = 20e3  # 20 km sub-aperture size.
+        sampling = 100.0  # 100 m/pixel.
```
deiruch, line 74: To reduce memory size, the resolution could be reduced.

deiruch:
@rohitcbscient I added some suggestions how to fix this easily, if it is really caused by memory issues in `test_ionosphere`.

karabo/test/test_ionosphere.py
Lukas113: As far as I know it is because the `Imager` requires that a `.ms` path exists (because the underlying libraries demand that)

deiruch: Created #438 to tackle this.

Lukas113:
I don't really know how to fix it. I tried to figure out where the error might come from by starting the test function as a script and not through a test-framework to get stderr and stdout. However, it doesn't really come to the point where something is printed (see [action](https://github.com/i4Ds/Karabo-Pipeline/actions/runs/5343319020/jobs/9686565709#step:7:125)) which doesn't make any sense to me. I put the prints after the imports and before the function call. The imports shouldn't do any harm because we have the same imports in other scripts as well.

To conclude, I don't know what the issue is. I just know it comes from the [karabo/test/test_ionosphere.py](https://github.com/i4Ds/Karabo-Pipeline/blob/326-implement-ionospheric-screen-into-karabo/karabo/test/test_ionosphere.py) because when I commented it out, all tests passed in previous attempts.

Lukas113:
The issue was that the time to create atmosphere phase screens using `ARatmospy.ARScreens.run` was too long (number of iterations) and therefore had memory requirement (more than 15GB) which was too large for the runners. To solve this issue, the time for the `run` function was reduced from 240 iterations (4h) to 1 iteration (1min). The memory requirement is then jsut 64MB.

Lukas113 (approved):

Issue #431, opened 2023-06-06
Add primary beam correction


jejestern:

jejestern:
Was closed by #432 and we will work on non-gaussian beams seperately.

PR #432, opened 2023-06-06
Js 423 include line emission simulation

jejestern:
Includes gaussian primary beam correction for one pointing (issue #431)

### DIFF
--- a/karabo/simulation/line_emission.py
+++ b/karabo/simulation/line_emission.py
@@ -555,7 +555,7 @@ def gaussian_fwhm_meerkat(freq: IntFloat) -> np.float64:
     :return: The power pattern FWHM of the MeerKAT telescope at this frequency in
              degrees.
     """
-    root = np.sqrt(89.5 * 86.2)
+    root = cast(np.float64, np.sqrt(89.5 * 86.2))
     gaussian_fwhm = root / 60.0 * (1e3 / (freq / 10**6))
 
     return gaussian_fwhm


karabo/simulation/line_emission.py
```diff
@@ -0,0 +1,639 @@
...
+    dirty_image = cast(NDArray[np.float_], dirty.data[0][0])
+
+    return dirty_image
+
+
+def oskar_imager(
+    outfile: str,
+    ra_deg: float = 20,
+    dec_deg: float = -30,
+    cut: float = 1.0,
```
Lukas113, line 131: A minor thing. If it's a user-api function, I recommend to be as less restrictive as possible when it comes to type-hints while remaining correctness. In this case, you specify that `ra_deg` and `dec_deg` must be of type `float`. But sometimes it is also fine if a user provides an `int`. So if a user want to just write 20 or provide a np.float_ instead of 20.0, he or she will get a type-error (if their IDE supports it). There are some pre-defined Union types for that case in `_types.py`. Here I'd recommend to use `IntFloat` of the mentioned module instead of just `float`.

In the end it isn't a big issue, but it's just a little more user-friendly and an easy fix. I won't mark all cases in this file, but I've seen more like this. I recommend you to adjust it wherever you find it reasonable.

karabo/simulation/sky_model.py
```diff
@@ -816,3 +839,84 @@ def get_random_poisson_disk_sky(
...
+        """
+
+        # Read in the catalog (only works for catalogs which are not very large)
+        catalog = h5py.File(path, "r")
+        print("The catalog keys are:", list(catalog.keys()))
+        print("The unit of the flux given here is:", catalog["Flux"].attrs["Units"])
+        print(
+            "Number of elements in the complete catalog:",
+            len(catalog["Declination"][()]),
+        )
```
Lukas113, line 891: I recommend to add a `verbose` option as an argument. As mentioned, I'm a fan of verbose functions if they take a while or if the output info is important to know. However, I also like the option to turn it off if needed.

Lukas113 (changes requested):

Lukas113 (approved):

PR #435, opened 2023-06-09
Replaced unittest and it's classes with pytest and functions

kenfus:


kenfus:
Did you update the .run_tests() function in utils?

.github/workflows/test.yml
```diff
@@ -46,9 +45,4 @@ jobs:
           conda activate test_karabo
           export IS_GITHUB_RUNNER=true
           pip install . --no-deps
-          python -m xmlrunner -o ./test-reports/junit
-      - name: Publish Unit Test Results
-        uses: EnricoMi/publish-unit-test-result-action@v1
-        if: always()
-        with:
-          files: test-reports/**/*.xml
+          pytest -s .
```
kenfus, line 48: I think -s causes too much spam in the log.

Lukas113: ok, I'll remove it

Lukas113: I just thought, since we have a lot of print-messages in our tests, it would be desired to get them. But now I think you're right. You can get the prints locally if you really want by adding the -s option.

Lukas113: done

Lukas113:
Do you think I can merge this branch now?

PR #440, opened 2023-06-21
Adds compute and closure of h5 connection when it's not needed anymore

kenfus:


karabo/simulation/sky_model.py
```diff
@@ -165,6 +168,7 @@ def __init__(
         sources: Optional[SkySourcesType] = None,
         wcs: Optional[WCS] = None,
         precision: Type[np.float_] = np.float64,
+        h5_file_connection: Optional[h5py.File] = None,
```
Lukas113, line 171: Please add argument docstring to `__init__`

karabo/simulation/sky_model.py
```diff
@@ -196,6 +201,46 @@ def __set_sky_xarr_dims(self, sources: SkySourcesType) -> None:
...
+        necessary to obtain the actual data and stores it in the `_sources` attribute.
+        After the computation is complete,
+        it calls the `close` method to close the connection to the HDF5 file.
+
+        Returns:
+        None
+        """
+        # Dask array inside the xarray to numpy array.
+        if self.sources is not None and isinstance(self.sources, xr.DataArray):
+            print("Computing the data array.")
```
Lukas113, line 240: I think `and isinstance(self.sources, xr.DataArray)` is unnecessary since `sources` should be of type `None | xr.DataArray`. 

In addition, if you think the print is useful information to a user when this method is called fine. But to me this is looking more like a debug-message and not a console-info message. However, if you think it should be there, just leave it.

karabo/simulation/sky_model.py
```diff
@@ -196,6 +201,46 @@ def __set_sky_xarr_dims(self, sources: SkySourcesType) -> None:
...
+            self._sources = self.sources.compute()
```
Lukas113, line 241: Here you access `self._sources` and not the property `sources`. I'm pretty sure this is not correct since the property takes care of some things and should ensure that `sources` is used correctly.

Please look at the property. And in case you're sure you don't want to use it, please leave a comment why you don't use and take care of things you have to change by yourself. If it needs to be cleared beforehand, just call `compute`, set `sources` to None and then set `sources` to what `compute` has returned.

Lukas113 (changes requested):

Lukas113 (approved):

PR #458, opened 2023-07-10
Js 452 make line emission simulation compatible with large input sky

jejestern:


karabo/simulation/line_emission.py
```diff
@@ -563,47 +569,53 @@ def line_emission_pointing(
...
     if os.path.exists(path_outfile):
         shutil.rmtree(path_outfile)
 
     os.makedirs(path_outfile)
 
+    # Load sky into memory and close connection to h5
+    sky.compute()
+
     if not client:
         "Print: Get dask client"
```
Lukas113, line 613: "Print: Get dask client" should be removed. A comment should start with a hashtag, not be a string. And since the next line of code already explains everything, this would be an unnecessary comment. If the intention is to print the message, then please do so. But I think this would not be a relevant print to the end-user?

Lukas113: In addition, line 624 `if client is not None:` doesn't seem to be necessary since 
```
if not client:
    "Print: Get dask client"
    client = DaskHandler.get_dask_client()
```

already narrows the type to be not None.

Lukas113 (changes requested):

Lukas113 (approved):

PR #460, opened 2023-07-12
added CaptureSpam to KaraboResource

Lukas113:
Added CaptureSpam to karabo_resource.py to be able to capture third-party library spam-messages.

Should be good enough. I tested it with all print-options (not logging). The only drawback is, that it can only check for each new-line if it has already been printed. So if e.g. a third-party library wraps their output in something like "--------------------\n" multiple times, it gets captured after the first time. It's also not able to capture with regex, just exact-match. But for the second case, CaptureSpam could be extended to support capturing by regex (I think).

### DIFF
--- a/karabo/karabo_resource.py
+++ b/karabo/karabo_resource.py
@@ -98,11 +98,17 @@ class HiddenPrints:
 class CaptureSpam:
     """Captures spam of `sys.stdout` or `sys.stderr`.
 
-    Captures spam-messages of an external library.
+    Captures exact-match spam-messages of an external library.
     It checks each new line (not an entire print-message with
      multiple multiple newlines), if it has already been printed once.
     Don't use CaptureSpam if the external library function
      provides a way to suppress their output.
+
+    Example usage:
+        ```
+        with CaptureSpam():
+            library_spam_fun()
+        ```
     """
 
     def __init__(self, stream: TextIO = sys.stdout):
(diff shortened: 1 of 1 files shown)


karabo/karabo_resource.py
```diff
@@ -87,3 +93,66 @@ def __exit__(
...
             sys.stderr = self._original_stderr
+
+
+class CaptureSpam:
+    """Captures spam of `sys.stdout` or `sys.stderr`.
+
+    Captures spam-messages of an external library.
+    It checks each new line (not an entire print-message with
+     multiple multiple newlines), if it has already been printed once.
+    Don't use CaptureSpam if the external library function
```
kenfus, line 104: Add an example on how to use it? Also, when should the user use it (e.g. when Oskar spams you).

Lukas113: Hmm, I thought the the usage is self-explanatory. Just use it like every `__enter__` `__exit__` function:

```python
with CaptureSpam():
    library_spam_fun()
```

When to use it is explained in the docstring. Do you want to change something there?

kenfus: Yes, a small example like that is already enough! As an example, when Oskar spams you with `WARNING: Gaussian ellipse solution failed for ...` that would be a good example to be captured! I don't think that it is self-explanatory.

Lukas113: Hmm, ok. Do you know are the messages exactly the same? Because currently just exact-matches are captured.

Lukas113: done

kenfus (approved):

PR #462, opened 2023-07-13
fixed OSKAR library not found in Dockerfile

Lukas113:
There was an issue using all our [docker-images](https://github.com/i4ds/Karabo-Pipeline/pkgs/container/karabo-pipeline) up until version `0.18.0`. `run_simulation` failed, where OSKAR python bindings claimed, that OSKAR library was not found (see image blow). This is essentially not true. They just failed an import of a sub-dependency. The issue was, that they captured their import error in `settings_tree.py`:

```python
try:
    from . import _settings_lib
    from . import _apps_lib
except ImportError:
    _settings_lib = None
    _apps_lib = None
```
There they try to import `_settings_lib.cpython-39-x86_64-linux-gnu.so` and `_apps_lib.cpython-39-x86_64-linux-gnu.so` which are their own .so files available in the oskar package location. The main issue is, that there they point to the exact cuda-runtime file `libcudart.so.11.0` (exact match for whatever reason) which comes with `cuda-toolkit 11.*` and didn't find it. This means, that the system must provide `cuda-toolkit 11.*` or OSKAR will fail when creating a settings-tree.

`cuda-toolkit 11.*`  was not available our docker-environment. I thought it must come as a dependency since we already have some cuda-toolkit dependencies in karabo like `cuda-cudart` or `libcufft`. However, it was ont available. I also tried to install the cuda toolkit manually using `conda install -c "nvidia/label/cuda-11.7.0" cuda-toolkit`. But it didn't work either. I tried to install the cuda-toolkit from the official [developer.nvidia](https://developer.nvidia.com/cuda-downloads?target_os=Linux&target_arch=x86_64&Distribution=Ubuntu&target_version=22.04&target_type=deb_local). There I got an error, that the glibc version `libc6 2.31-13+deb11u3` version of the previous used base-image [continuumio/miniconda3](https://hub.docker.com/r/continuumio/miniconda3) was too low, because to install the nvidia-toolkit 11.*, at least `libc6 >=2.34` is required. Even their latest base image didn't reach the glibc requirement. I assume that this is the issue, why the cuda-toolkit wasn't successfully installed. But not sure though, I don't have experience for this kind of stuff.

So in the end, what worked was taking the base-image `nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04` which had the nvidia-toolkit installed and the required file was there. Also the glibc requirement (`libc6 = 2.35`) was met.  The code shown in the image-below succeeded in the new build. So I think it is fine to include it as it is to the main-branch (even though the images are still not fully tested).

image

### DIFF
--- a/docker/user/Dockerfile
+++ b/docker/user/Dockerfile
@@ -1,33 +1,24 @@
 # Create build container to not have copied filed in real container afterwards
-FROM --platform=amd64 continuumio/miniconda3:4.12.0 as build
+FROM nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04 as build
 ARG KARABO_TAG
 RUN apt-get update && apt-get install -y git
 RUN git clone --branch ${KARABO_TAG} --depth=1 https://github.com/i4Ds/Karabo-Pipeline.git
 
-FROM --platform=amd64 continuumio/miniconda3:4.12.0
+FROM nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04
 ARG KARABO_TAG
+RUN apt-get update && apt-get install -y libarchive13 wget curl nano
+RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-py39_23.5.0-3-Linux-x86_64.sh -O ~/miniconda.sh && \
+    /bin/bash ~/miniconda.sh -b -p /opt/conda
+ENV PATH=/opt/conda/bin:$PATH
+RUN conda init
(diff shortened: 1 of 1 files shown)


Lukas113:
And to clarify, on my local machine with karabo installed, `libcudart.so.11.0` is in `~/miniconda3/envs/karabo_dev_env/lib/libcudart.so.11.0`, thus available through my virtual environment. I have also a glibc version of 2.35. Thus, I think we shouldn't update our direct dependencies. I assume that Karabo doesn't run on systems with an old glibc library for that reason.

PR #479, opened 2023-07-24
Vt add mgcls

kenfus:
- Adds MGCLS data with some description.
- Files are probably too big to make any tests yet. Maybe later?
- Improved doc in how to upload to CSCS
- Added new functionality to the data download.


karabo/data/external_data.py
```diff
@@ -115,3 +125,52 @@ def __init__(self) -> None:
...
+
+
+class MGCLSFilePaths(ContainerContents):
+    def __init__(self, regexr_pattern: str) -> None:
+        super().__init__(f"MGCLS/{regexr_pattern}")
+
+
+class MGCLSFitsGzDownloadObject(FitsGzDownloadObject):
+    def __init__(self, file_name: str) -> None:
+        super().__init__(file_name, None)
```
Lukas113, line 176: To me this class seems odd. Firstly, there is multiple inheritance of `DownloadObject` through`FitsGzDownloadObject` and I don't see a reason why. Additionaly, I can't see what this class has to do with `MGCLSFitsGz` since no parameters of this class (also in combination with it's superclass) have to do anything with the MGCLS files. They're just general classes as far as I see.

Maybe I'm missing something?

karabo/simulation/sky_model.py
```diff
@@ -1225,6 +1229,82 @@ def get_GLEAM_Sky(frequencies: Optional[List[GLEAM_freq]] = None) -> SkyModel:
...
+        -------
+        List[Image]
+            List of images from the MGCLS Enhanced Products bucket.
+
+        Example:
+        --------
+        >>> mgcls_files = SkyModel.get_MGCLS_data('Abell_(?:2744)_.+_I_.+')
+        >>> print(mgcls_files)
+        <Xarray.Image frequency: 5, polarisation: 1, y: 3617, x: 3617>
+        >>> filtered = mgcls_files[0].sel({'frequency': mgcls_files[0].frequency[0], 'polarisation': 'I'}) # type: ignore[index] # noqa: E501
```
Lukas113, line 1286: I think we don't need `# type: ignore[index] # noqa: E501` in docstrings.

karabo/simulation/sky_model.py
```diff
@@ -1225,6 +1229,82 @@ def get_GLEAM_Sky(frequencies: Optional[List[GLEAM_freq]] = None) -> SkyModel:
...
+        >>> filtered['pixels'].plot.imshow()
+        <matplotlib.image.AxesImage object at 0x7f8a1545fc10>
+        """
+        # Get Paths
+        mfg = MGCLSFilePaths(regexr_pattern=regex_pattern)
+        file_paths = mfg.get_file_paths()
+        if len(file_paths) == 0:
+            raise KaraboSkyModelError("No files found for the provided regex pattern.")
+        # Get the files
+        mgcls_files = []
```
Lukas113, line 1296: Seems strange to me why `mypy` (almost `--strict`) doesn't complain about something like this. Usually, when you declare an empty list, it doesn't have a generic type-hint and mypy should blame you that you are returning `list` instead of `List[Image]`. However, I think we should do it regardless what mypy says and therefore just change it to `mgcls_files: List[Image] = []`.

Lukas113 (changes requested):

karabo/simulation/sky_model.py
kenfus: The problem was, if I write multicode here, flake8 says things like "invalid code", even though it is valid. So I just ignore line length.

karabo/data/external_data.py
```diff
@@ -36,13 +38,21 @@ def __init__(
...
+            os.makedirs(os.path.dirname(self.path), exist_ok=True)
+            with open(self.path, "wb") as file:
+                for chunk in response.iter_content(
+                    chunk_size=8192
+                ):  # Download in 8KB chunks
+                    file.write(chunk)
+        except BaseException:
+            # Remove the file if the download is interrupted
+            if os.path.exists(self.path):
+                os.remove(self.path)
```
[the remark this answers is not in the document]
kenfus: I think this is always a file. You mean when e.g. we download measurements-set, directly as a folder?

karabo/simulation/sky_model.py
Lukas113: Ok, weird. However, not really an issue to keep it as it is.

karabo/data/external_data.py
```diff
@@ -115,3 +125,52 @@ def __init__(self) -> None:
             "https://object.cscs.ch/v1/AUTH_1e1ed97536cf4e8f9e214c7ca2700d62"
             + "/karabo_public/example_map.h5",
         )
+
+
+class FitsGzDownloadObject(DownloadObject):
+    def __init__(self, file_name: str, folder_name: Optional[str]) -> None:
```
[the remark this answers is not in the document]
kenfus: I restructured it a bit.

karabo/data/external_data.py
```diff
@@ -115,3 +125,52 @@ def __init__(self) -> None:
...
+            file_name,
+            f"{base_path}/{file_name}",
+        )
+
+
+class ContainerContents:
+    def __init__(self, regexr_pattern: str):
+        self.regexr_pattern = regexr_pattern
+
+    def get_container_content(self) -> str:
```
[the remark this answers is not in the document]
kenfus: I think I leave it like this and add some documentation for the user to use.

karabo/data/external_data.py
```diff
@@ -115,3 +125,52 @@ def __init__(self) -> None:
...
+        response.raise_for_status()
+
+        # Get the XML content as a string
+        return response.text
+
+    def get_file_paths(self) -> List[str]:
+        xml_content = self.get_container_content()
+        url_pattern = re.compile(self.regexr_pattern)
+        urls = url_pattern.findall(xml_content)
+        return urls
```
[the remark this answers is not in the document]
kenfus: I tried different things, like XML parser. However, I was not able to make it work correctly. If you want to have a go at it, feel free!

karabo/data/external_data.py
```diff
@@ -115,3 +125,52 @@ def __init__(self) -> None:
...
+
+
+class FitsGzDownloadObject(DownloadObject):
+    def __init__(self, file_name: str, folder_name: Optional[str]) -> None:
+        base_path = (
+            "https://object.cscs.ch/v1/AUTH_1e1ed97536cf4e8f9e214c7ca2700d62/"
+            "karabo_public"
+        )
+        if folder_name is not None:
+            base_path += f"/karabo_public/{folder_name}"
```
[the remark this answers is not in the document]
kenfus: Yes, it's restructured

karabo/data/external_data.py
```diff
@@ -19,42 +21,51 @@ def valida_cache_directory_exists() -> None:
 
     @staticmethod
     def get_cache_directory() -> str:
-        cache_path = f"{KaraboCache.base_path}/karabo_cache"
+        cache_path = f"{KaraboCache.base_path}{os.sep}karabo_cache"
```
Lukas113, line 24: `os.path.join(KaraboCache.base_path, "karabo_cache")` is looking cleaner to me because we are use it everywhere in our code and I've seen a lot of examples with join on third-party code-bases, but never in this style.

karabo/data/external_data.py
```diff
@@ -19,42 +21,51 @@ def valida_cache_directory_exists() -> None:
...
 
 
 class DownloadObject:
     def __init__(
         self,
-        name: str,
-        url: str,
+        file_path: str,  # e.g. "karabo_public/point_sources_OSKAR1_battye.h5"
+        base_url: str = "https://object.cscs.ch/v1/"
+        + "AUTH_1e1ed97536cf4e8f9e214c7ca2700d62",
```
Lukas113, line 33: I think it's preferable to make something like:
- `karabo_cscs_object_storage = "https://object.cscs.ch/v1/AUTH_1e1ed97536cf4e8f9e214c7ca2700d62`
- `karabo_public = "karabo_public"`
global scoped. The reason is to have a more organized download-variable structure. In addition, thee also would be importable from other modules if needed.

Then you just can set it's default values accordingly.

karabo/data/external_data.py
```diff
@@ -19,42 +21,51 @@ def valida_cache_directory_exists() -> None:
...
+            # Remove the file if the download is interrupted
+            if os.path.exists(self.local_path):
+                os.remove(self.local_path)
+            raise
 
     def __is_downloaded(self) -> bool:
-        if os.path.exists(self.path):
+        if os.path.exists(self.local_path):
             return True
         return False
```
Lukas113, line 61: change to: `return os.path.exists(self.local_path)` since simple if-else boolean returns are considered as an anti-pattern.

karabo/data/external_data.py
```diff
@@ -19,42 +21,51 @@ def valida_cache_directory_exists() -> None:
...
 
     def get(self) -> str:
         if not self.__is_downloaded():
-            print(f"{self.name} is not downloaded yet.")
-            print("Downloading and caching for future uses to " f"{self.path} ...")
+            print(f"{self.local_path.split(os.sep)[-1]} is not downloaded yet.")
```
Lukas113, line 65: Did you mean `self.url` here instead of `local_path.split(os.sep)[-1]`? Otherwise it's kind of meaningless in my opinion since you already print `self.local_path` in the print-statement below.

karabo/data/external_data.py
```diff
@@ -66,52 +77,91 @@ def is_available(self) -> bool:
...
+        return resp.status_code == 206
 
 
 class GLEAMSurveyDownloadObject(DownloadObject):
     def __init__(self) -> None:
         super().__init__(
-            "GLEAM_ECG.fits",
-            "https://object.cscs.ch/v1/AUTH_1e1ed97536cf4e8f9e214c7ca2700d62"
-            + "/karabo_public/GLEAM_EGC.fits",
+            "karabo_public/GLEAM_EGC.fits",
```
Lukas113, line 86: As mentioned above, once `karabo_public` would be a global-spaced variable, it would be better to use it then and change it everywhere to: `f"{karabo_public}/my_fits_file.fits"`. This would reduce the number of magic substrings.

Lukas113 (changes requested):

karabo/data/external_data.py
kenfus: No, It's just a nicer print statement. One what is not downladed and the second one where it will be downloaded.

Lukas113 (approved):

PR #482, opened 2023-07-27
Improve paths used in line emission

lmachadopolettivalle:
Make file paths used in `line_emission.py` backwards compatible and clearer to the end user.

Before, the output files were split between the requested directory and a subdirectory within it. Now, all result files are saved under the folder path selected by the user.

Also, changed the test and the example `ipynb` to use the new function signatures consistently.


Lukas113 (approved):
LGTM

karabo/simulation/line_emission.py
```diff
@@ -748,7 +743,7 @@ def simple_gaussian_beam_correction(
     dirty_image_corrected = dirty_image / beam
 
     fits.writeto(
-        path_outfile / "test_line_emission_GaussianBeam_Corrected.fits",
+        path_outfile / "line_emission_total_image_beamcorrected.fits",
```
jejestern, line 746: Change path_outfile to outpath and it would be great to add a docstring

jejestern (changes requested):

karabo/simulation/line_emission.py
```diff
@@ -720,30 +720,42 @@ def gaussian_beam(
...
     img_size: int = 4096,
 ) -> Tuple[NDArray[np.float_], fits.header.Header]:
-    path_outfile = Path(path_outfile)
+    """
+    Apply Gaussian Beam correction to the Dirty Image.
+
+    :param outpath: Path where beam data and output image will be saved.
+    :param dirty_image: Dirty Image, e.g. output from line_emission_pointing.
+    :param gaussian_fwhm: FWHM of the Gaussian in degrees.
+    :param ra_deg: Right ascension coordinate of center of Gaussian.
```
jejestern, line 737: This is is the phase center right ascension and for the declination as well. Due to that fact it is also the center of the Gaussian. But the definition is that it is the phase center

jejestern (changes requested):

jejestern (approved):

PR #483, opened 2023-07-27
Lg 472 code coverage

### LINKED ISSUE #472: Coverage reports of all Unit Tests
Our unit tests should report coverage.

Lukas113:
closes #472


.github/workflows/test.yml
```diff
@@ -4,22 +4,31 @@ on:
   schedule: # Every night at 5am
     - cron: '0 5 * * *'
   pull_request:
+    types: [opened, synchronize, reopened]
   workflow_dispatch:
+    inputs:
+      run_slow_tests:
```
kenfus, line 10: I think variables are case-sensitive, no? Or does this still work?

.github/workflows/test.yml
```diff
@@ -4,22 +4,31 @@ on:
...
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
       - name: Install Conda
         uses: conda-incubator/setup-miniconda@v2
         with:
           miniforge-variant: Mambaforge
       - name: Install Deps
         shell: bash -el {0}
         run: |
-          mamba create -n test_karabo
```
kenfus, line 22: Why back to conda?

kenfus (changes requested):

.github/workflows/test.yml
Lukas113: No, it's a workflow input. But still, it's a legacy feature which I'll remove since it has not meaning in the current workflow. I'll remove in the next commit.

.github/workflows/test.yml
Lukas113: Because in my experience using mamba for everything is weird and it's just needed for dependency resolving purpose. So I thought either do everything with mamba (which is the same as with conda, except for deps-resolving), or just the part that really matters.

But in this case we should probably do everything with mamba even though it shouldn't matter at all.

Lukas113: Changed back because mamba weirdly needs an `init` before calling activate, but other mamba functions work. So I just let mamba do dep-resolving and leave everything else to conda.

image

kenfus (approved):

PR #491, opened 2023-08-02
490 fix notebooks

### LINKED ISSUE #490: Notebooks are fixed for testing
Currently, all Notebook tests are failing (see [action](https://github.com/i4Ds/Karabo-Pipeline/actions/runs/5715747385/job/15485757597)). This needs to be fixed as soon as possible.

image

Lukas113:
closes #490


Lukas113:
Converted to a draft because it works locally but not on github runners and they're just triggered after an opened PR.

When everything works, then
```
if [[ ${{ github.event_name }} == pull_request ]]; then
    export RUN_SLOW_TESTS=true
```
needs to be changed back to `false`

Lukas113:
I've disabled the notebook-tests entirely because they don't run on the github runners (see [action](https://github.com/i4Ds/Karabo-Pipeline/actions/runs/5737524192/job/15549330638)). I'm not entirely sure about the reason, maybe because of limited RAM of 7GB?

However, I don't see how this can be fixed easily. In the meantime, I suggest running the notebooks on your local machine by setting `RUN_SLOW_TESTS=true` should be good enough.

kenfus:
Hmm... That is wierd. Maybe check which one fails? Is that possible? In the notebook `HIIM_Img_Recovery` there is also leftovers of some dask tests, I guess? That is very annoying if our notebooks are not checked regularly

Lukas113:
Ok, just the `source_detection.ipynb` is causing the issue. Therefore, the other notebooks are run. I also removed `RUN_SLOW_TESTS` option since the running two notebooks don't take a long for execution.

Lukas113:
I've just removed the other two notebooks. They did run on individually on the runners, however they didn't in the whole test-framework (see [action](https://github.com/i4Ds/Karabo-Pipeline/actions/runs/5750283400)).

Lukas113:
So, after a google-searched I figured out that the issue is probably that the maximum disk-space is at max (10GB). This is probably caused because our messy behavior of `FileHandle` and it's usage. As far as I know there are cases where multiple copies of the same file are created on different locations. In addition, there is no easy way to clean-up these automatically intermediate files. Therefore, the disk-usage just grows and comes at it's limit as some point. This has to change if we want to keep running a lot of tests on the GitHub runners.

Lukas113:
In the meantime, two Notebooks which did run on the runners, but not together with the other tests are skipped until we fix this issue.

kenfus (approved):

PR #492, opened 2023-08-02
Added transpose to get_pixel_position_of_sources()

kenfus:
- Also removed useles test.


Lukas113 (approved):
LGTM

I'd prefer to merge this after #491 is merged successfully. Because then the tests on the main-branch on a merged PR should also be successful. Otherwise they fail and the Codecov won't get submitted.

Lukas113:
In addition, I believe `codecov/project` failed because there is not coverage-report at the latest commit on the main branch. But because the "error message" in Details doesn't show any meaningful information why it fails I'm not sure about it. I also haven't found something useful after a google-search.

So I just hope that after #491 is merged successfully it will work. Otherwise I must have a look at this issue.

Issue #433, opened 2023-06-08
Line emission: Equally spaced frequency or redshift bins


jejestern:
Line emission: Have equally spaced frequency bins instead of redshift bins, but keep having equally spaced redshift bins as a possiblity

PR #493, opened 2023-08-02
Line Emission: Default to equally spaced bins in frequency space

lmachadopolettivalle:
Per #433 add support for channels equally spaced in frequency space, and use this as the default. Also, added tests for the conversions between frequency and redshift for the 21cm line.


karabo/simulation/line_emission.py
```diff
@@ -256,53 +256,72 @@ def sky_slice(sky: SkyModel, z_min: np.float_, z_max: np.float_) -> SkyModel:
...
 
-def redshift_slices(
-    redshift_obs: Union[NDArray[np.float_], xr.DataArray],
-    channel_num: int = 10,
-) -> NDArray[np.float_]:
-    """
-    Creation of the redshift bins used for the line emission simulation based on the
-    observed redshift range from the sky.
+def convert_z_to_frequency(z):
+    """Turn given redshift into corresponding frequency (Hz) for 21cm emission."""
```
jejestern, line 261: Please add :param for z and :return, also what type is z?

jejestern (changes requested):

jejestern (approved):

karabo/simulation/line_emission.py
```diff
@@ -256,53 +256,80 @@ def sky_slice(sky: SkyModel, z_min: np.float_, z_max: np.float_) -> SkyModel:
...
+) -> Union[NDArray[np.float_], xr.DataArray, IntFloat]:
+    """Turn given frequency (Hz) into corresponding redshift for 21cm emission.
+
+    :param freq: Frequency values to be converted into redshifts.
+
+    :return: Redshifts corresponding to input frequencies.
+    """
 
-    return redshift_channel
+    return (c.value / (0.21 * freq)) - 1
```
Lukas113, line 282: A Union return is often an issue for mypy. To me it seems like you could solve this with `@overload` (examples see other code in Karabo) which would still be correct.

karabo/simulation/line_emission.py
```diff
@@ -256,53 +256,80 @@ def sky_slice(sky: SkyModel, z_min: np.float_, z_max: np.float_) -> SkyModel:
...
-    :return: The channels of the observed redshift range.
+    :return: Frequencies corresponding to input redshifts.
     """
-    print("Smallest redshift:", np.amin(redshift_obs))
-    print("Largest redshift:", np.amax(redshift_obs))
 
-    redshift_channel = np.linspace(
-        np.amin(redshift_obs), np.amax(redshift_obs), channel_num + 1
-    )
+    return c.value / (0.21 * (1 + z))
```
Lukas113, line 269: Same issue as below. I think using `@overload` should be better for type-hinting. However, make sure to test the type-hints on runtime so that they're correct.

Lukas113 (changes requested):
LGTM apart from the 2 changes I've requested.

PR #495, opened 2023-08-02
Test for filter by flux exists

kenfus:
- adds test for filter by flux, for both h5 skys and fits skies.


karabo/test/test_skymodel.py
```diff
@@ -78,6 +78,30 @@ def test_filter_sky_model_h5():
...
+    assert np.all((filtered_sky[:, 2] >= 46.872) & (filtered_sky[:, 2] <= 300.0))
+
+
+def test_filter_flux_sky_model_h5():
+    sky = SkyModel.get_BATTYE_sky(which="diluted")
+    assert sky[:, 2].max() > 1.15626285e-26, "Test data is not correct"
+    filtered_sky = sky.filter_by_flux(9.95626285e-27, 1.15626285e-26)
+    assert np.all(
+        (filtered_sky[:, 2] >= 9.95626285e-27) & (filtered_sky[:, 2] <= 1.15626285e-26)
+    )
```
Lukas113, line 102: I don't think a second filter-test-function does improve the tests. If you think otherwise, please explain.

kenfus: I think it does, the input is different. We should Parameterized this but I mean the Flux values are very different. It's good to know that it works with different skies.

Lukas113: Ok, fair point. But for that reason features like `pytest.mark.parametrize` exist. Therefore, having multiple tests which are testing the same thing doesn't seem necessary.

In addition, I think since the purpose of the test is to test `filter_by_flux`, it would be better to have a defined (meaning hard-coded) sky as input, where you are certain that after applying a filter function how the sky must look like. So I suggest replacing the "blackbox" sky-models with a simpler sky-model for this test.

kenfus: I disagree, I think this is OK. I don't think we will ever change gleam sky or the diluted sky. So we also test functions in the way a user would test them.

Lukas113 (approved):

PR #497, opened 2023-08-02
Adds aratmospy to the user package

kenfus:
- Adds testing of user package
- Adds missing package aratmospy


conda/meta.yaml
```diff
@@ -19,6 +19,7 @@ requirements:
     - pip
   run:
     - python=3.9
+    - aratmospy=1.0.0
     - oskarpy=2.8.3
     - rascil=1.0.0
     - numpy=1.22
```
Lukas113, line 25: I totally forgot to update the dependencies in the meta.yaml last time I've updated our direct dependencies.

However, please also update the rest of the conda.yaml dependencies so that is has the same state as the `environment.yml`.

Lukas113 (changes requested):

Lukas113 (approved):

PR #499, opened 2023-08-04
Js 424 include coadd classfunction

jejestern:


karabo/examples/Mosaicking_continuous.ipynb
lmachadopolettivalle, line 1: A few suggestions:
- Cell 2: `t_obs` is March 20, 2020, but the comment says September 1st, 2022 - should we remove the comment or change the variable value?
- Cell 3: Could we better explain the variables and their use cases? Also, a few of them seem redundant. E.g. `location` vs `phase_center` vs `ra_deg` and `dec_deg`, and `size` vs `area` vs `outer_rad`
- Cell 7: We could add a reference to the equation for the FWHM_real calculation, as well as to why 0.87 is the ideal value for `fac`. Or make this into a function, where we explain the specific values being used.
- Cell 7: `pointings = [center1, center 2]` would be more straightforward
-  Cell 10: is the resolution always 5 in the mosaic_header, or should this be determined from other variables in the code?

karabo/imaging/mosaic.py
```diff
@@ -0,0 +1,125 @@
+import os
+import shutil
+from pathlib import Path
+from typing import Optional
+
+from MontagePy.archive import *  # noqa
```
lmachadopolettivalle, line 6: It'd be more readable to explicitly import what we need instead of importing *

karabo/imaging/mosaic.py
```diff
@@ -0,0 +1,125 @@
...
+from MontagePy.main import mAdd, mHdr, mImgtbl, mProjExec
+
+from karabo.util._types import FilePathType
+
+
+def mosaic_directories(output_directory_path: FilePathType) -> None:
```
lmachadopolettivalle, line 12: In the docstring it would be helpful to add what the folder structure will be.

The directory structure can be generated using e.g. https://stackoverflow.com/a/3455651

karabo/imaging/mosaic.py
```diff
@@ -0,0 +1,125 @@
...
+    Creates the header for the mosaic you want to create.
+
+    :param output_directory_path: The folder path in which all subfolders used for the
+                                  mosaic are.
+    :param location: The center of the mosaic in "RA, Dec" in degrees (Equatorial).
+    :param width: Image width in degrees.
+    :param height: Image height in degrees.
+    :param resolution: Image pixel resolution (in arcsec).
+    :param sin_projection: Set to true if you want the output mosaic to be in SIN
+                           projection. The output fits files from a karabo simulation of
```
lmachadopolettivalle, line 50: The description of `sin_projection` could be more clear. Is it recommended to set it as `True`, since the Karabo output files also use this projection?

karabo/test/test_mosaic.py
```diff
@@ -0,0 +1,187 @@
...
+    DilutedBATTYESurveyDownloadObject,
+    SingleFileDownloadObject,
+    cscs_karabo_public_base_url,
+)
+from karabo.imaging.mosaic import mosaic, mosaic_directories, mosaic_header
+from karabo.simulation.line_emission import freq_channels, karabo_reconstruction
+from karabo.simulation.sky_model import SkyModel
+
+
+# DownloadObject instances used to download different golden files:
```
lmachadopolettivalle, line 19: Update this comment to reflect the files being used here. Also, explain the difference between the uncorrected and the uncorrected area files

karabo/test/test_mosaic.py
```diff
@@ -0,0 +1,187 @@
...
+
+    # Setting for pointings
+    img_size = 1024
+    fac = 0.87
+    FWHM_real = np.sqrt(89.5 * 86.2) / 60.0 * (1e3 / (freq_mid / 10**6)) * u.deg
+    size_pntg = 1.2 * FWHM_real
+    offset = FWHM_real * fac
+    center1 = coords.SkyCoord(ra=ra * u.deg, dec=dec * u.deg, frame="icrs")
+    center2 = coords.SkyCoord(ra=ra * u.deg + offset, dec=dec * u.deg, frame="icrs")
+    pointings = [
```
lmachadopolettivalle, line 98: Same as in the `ipynb`, `pointings = [center1, center2]`

karabo/test/test_mosaic.py
```diff
@@ -0,0 +1,187 @@
...
+        center1,
+    ] + [center2]
+
+    # Settings for mosaic
+    location = "20.84, -30.0"
```
lmachadopolettivalle, line 103: Add a comment explaining why this specific choice of location? In the `ipynb` the location is set to be equal to the phase center, so it'd be useful to understand the difference

karabo/test/test_mosaic.py
```diff
@@ -0,0 +1,187 @@
...
+    size_w = 4.0 * u.deg
+    size_h = 2.5 * u.deg
```
lmachadopolettivalle, line 105: It would be helpful to add a comment on the meaning of these values, and specially why they should be different

lmachadopolettivalle (changes requested):

lmachadopolettivalle (approved):
LGTM 💯

Lukas113 (approved):
:thumbsup:

Issue #369, opened 2023-03-23
Testing Notebooks on GitHub runners work


Lukas113:
Currently, there is an unknown error. The notebook code started, but did not finish. The error output was: `Error: The operation was canceled`. I think this should be fixed to ensure that our notebooks are up to date.

The notebooks can be tested in the local test-setup. But even there, the notebook `HIIM_Img_Recovery.ipynb` is excluded, because even after 5 hours it did not finish. I think if possible, this should either be adapted to a shorter running time or once the notebooks can be tested on the runners tested there during the night.

Lukas113:
I've an update on this issue. I did some work in the PR #491 which addresses the issue #490 . The notebook `source_detection.ipynb` get's an `Error: Process completed with exit code 143` message (basically a gentle `kill` instead of `kill -9`) where I don't have an idea how to solve this issue.

The other notebooks however got an `System.IO.IOException: No space left on device` error (details are described in the PR #491 ). To solve this, we should change the usage of `FileHandle` in Karabo. This means that if a path of the intermediate file-products is provided, then there shouldn't be some files or directories created through `FileHandle`. In addition, multiple copies of the same file or directory shouldn't be created for no reason.

Therfore, to solve the disk-memory issue, I suggest a refactoring of `FileHandle` and it's usage with the following specifications:
- Just use `FileHandle` in case no path for the intermediate products is provided
- `FileHandle` should create a `prefix_{shortened(uuid4)}` directory inside a root-directory where the root is a static `DirPathType` and the `prefix` is provided through `__init__`.
- The `FileHandle` prefixed-directory path should be used inside the classes as an `str` or `DirPathType`, and not the `FileHandle` object itself. The reason is that other processes (outside of the class which created a FileHandle path) therefore don't have to care if the path comes from a user or from`FileHandle`. So don't use something like an `Union[FileHandle, str]` in the API signature.
- If you need to save a `FileHandle` object inside a class for some reason, save it as a protected or private variable. Therefore, the path access should be through a public or protected `DirPathType` variable.
- For clearing up, we should provide an instance-bound function which removes the directory bound to the instance (the prefix-directory) and a static function which removes the root directory of `FileHandle`.
- I think we should encouraged to remove the temporal directories wherever it makes sense (provided they got set through `FileHandle` and not through a custom path).

PR #503, opened 2023-08-07
369 test notebooks

Lukas113:
- Complete refactoring of `FileHandle` according to #369 .
- Improved some type-hints and docstrings
- Improved os-path-handling


Lukas113:
TODO: adapting tests, bugifxing according to pytest

I hope it doesn't brake something because there were a lot of changes. Even if the thest succeed, we don't know for sure since the coverage is not that high. However, I think once this PR is accepted, we should have a much better file- and dir-management than we had before.Then, we finally can remove from Karabo created directories and files safely in a convenient way.

karabo/simulation/interferometer.py
```diff
@@ -372,11 +407,21 @@ def __run_simulation_parallized_observation(
...
+        ms_dir = os.path.join(fh.subdir, "measurements")
+        os.makedirs(ms_dir, exist_ok=True)
+        vis_dir = os.path.join(fh.subdir, "visibilities")
+        os.makedirs(vis_dir, exist_ok=True)
         for observation_params in observations:
-            # Create params
-            interferometer_params = (
-                self.__create_interferometer_params_with_random_paths(input_telpath)
+            start_freq = observation_params["observation"]["start_frequency_hz"]
+            ms_file_path = os.path.join(ms_dir, f"start_freq_{start_freq}.MS")
```
kenfus, line 419: Nice!

karabo/simulation/interferometer.py
```diff
@@ -549,9 +578,13 @@ def __run_simulation_long(
                 pb.save_cst_file(beam[4], telescope=telescope)
                 pb.fit_elements(telescope)
 
-            # Create params
-            interferometer_params = (
-                self.__create_interferometer_params_with_random_paths(input_telpath)
+            start_freq = observation_params["observation"]["start_frequency_hz"]
```
kenfus, line 581: I think in observation long the start_frequency does not change, only the date. So the name should contain the current date.

karabo/simulation/pinocchio.py
```diff
@@ -73,8 +73,12 @@ def __init__(self, working_dir: Optional[str] = None) -> None:
         """
         if working_dir is not None:
             working_dir = os.path.abspath(working_dir)
-        self.wd = FileHandle(dir=working_dir)
-
+            self.wd = working_dir
+        else:
+            fh = FileHandler.get_file_handler(
```
kenfus, line 78: Here it warns me that no test is included. Maybe add a small one for `get_file_handler` ? Just assert that it gives back the correct path or something.

karabo/simulation/telescope.py
```diff
@@ -196,27 +197,31 @@ def get_OSKAR_telescope(self) -> OskarTelescope:
...
+        self.__write_position_txt(os.path.join(dir, "position.txt"))
         self.__write_layout_txt(
-            f"{dir}/layout.txt", [station.position for station in self.stations]
+            os.path.join(dir, "layout.txt"),
+            [station.position for station in self.stations],
         )
         i = 0
         for station in self.stations:
-            station_path = f"{dir}/station{'{:03d}'.format(i)}"
+            station_path = f"{dir}{os.path.sep}station{'{:03d}'.format(i)}"
```
kenfus, line 219: Unsure, but maybe you should use {station} somewhere in this path?

kenfus: Ah I see, we use `i` instead of station.

karabo/simulation/visibility.py
```diff
@@ -37,31 +37,49 @@ def __init__(
...
-            vis = Visibility(path=path)
+            raise ValueError(f"File must be a .ms or .vis file, but is {path=} instead")
         return vis
 
     @staticmethod
-    def is_measurement_set(path: str) -> bool:
-        return path.endswith(".ms") or path.endswith(".MS")
+    def is_measurement_set(path: DirPathType) -> bool:
+        path_ = str(path)
+        return path_.endswith(".ms") or path_.endswith(".MS")
```
kenfus, line 77: use .lower() ?

karabo/test/test_filehandler.py
```diff
@@ -0,0 +1,79 @@
...
+        FileHandler.root = tmpdir
+
+        # add 2 dirs created through FileHanlder with and without random content
+        fh1 = FileHandler(prefix="my_domain", verbose=True)
+        with open(os.path.join(fh1.subdir, "my_json.json"), "w") as outfile1:
+            json.dump({"A": "B"}, outfile1)
+        _ = FileHandler(prefix="my_other_domain", verbose=False)
+        assert len(os.listdir(tmpdir)) == 2
+
+        # create 3 additional random other dirs and files with and without content
```
kenfus, line 23: I see here that you create 4 files, no? Not 3.

karabo/util/file_handler.py
```diff
@@ -0,0 +1,186 @@
...
+        verbose: bool = True,
+    ) -> None:
+        """Creates `FileHandler` instance with the according sub-directory.
+
+        Args:
+            prefix: Prefix for easier identification of sub-directory.
+            verbose: Subdir creation and removal verbose?
+        """
+        self.verbose = verbose
+        subdir_name = str(uuid.uuid4())[:8]
```
kenfus, line 53: https://devblogs.microsoft.com/oldnewthing/20080627-00/?p=21823

not sure if we care but I'm still posting it.

karabo/util/file_handler.py
```diff
@@ -0,0 +1,186 @@
...
+            )
+        os.makedirs(self.subdir, exist_ok=False)
+
+    def clean_up(self) -> None:
+        """Removes instance-bound `self.subdir`."""
+        if os.path.exists(self.subdir):
+            if self.verbose:
+                print(f"Removing {self.subdir}")
+            shutil.rmtree(self.subdir)
+            if len(os.listdir(FileHandler.root)) == 0:
```
kenfus, line 75: Nice!

karabo/util/file_handler.py
```diff
@@ -0,0 +1,186 @@
...
+                        shutil.rmtree(path=path)
+                if len(os.listdir(FileHandler.root)) > 0:
+                    if verbose:
+                        print(
+                            f"`clean_up_fh_root` is not able safely remove "
+                            f"{FileHandler.root} because there are directories which "
+                            f"don't match {FileHandler.fh_dir_identifier=} or files."
+                        )
+                else:  # remove fh-root if dir is empty
+                    shutil.rmtree(FileHandler.root)
```
kenfus, line 140: CAn you not call `clean_up_fh_root` here?

kenfus (changes requested):

karabo/simulation/interferometer.py
Lukas113: You're right, thanks.

karabo/simulation/pinocchio.py
Lukas113: `get_file_handler` is tested in `test_file_handler.test_get_file_handler` that it does the right thing. Therefore, I disagree that we should do a test which checks if `FileHandler.get_file_handler` is working correctly there.

karabo/test/test_filehandler.py
Lukas113: Hmm, maybe I was not specific with the comment. What I meant is I created 3 additional dirs or files on the `tmpdir` level. I'll adapt this comment.

karabo/util/file_handler.py
Lukas113: Yeah you're right. Not a good enough reason to shorten the uuid just because it looks nicer when it's short at the cost of a potential error (even though it is small).

karabo/util/file_handler.py
Lukas113: Technically you can if you set `force` to True. However, I don't see a reason why you should call the function recursively which could lead to some unexpected behavior in the future.

kenfus (approved):

PR #516, opened 2023-09-19
Vt 514 dask usage via karabo is pit of success

kenfus:
- Added variable inside docker image so that we can dynamically create dask client
- Added back dask mpi
- Replaced some code for parallelization.


Lukas113 (approved):
LGTM

PR #520, opened 2023-10-10
Improve documentation for Line Emission

lmachadopolettivalle:
* Improve documentation in line emission example notebook
* Rerun script to verify it runs end to end


deiruch (changes requested):
We should not skip tests, especially not this one.

The idea of this test, was to have a very simple PINOCCHIO run, small enough that it can complete on GH runners. If it's still to slow, we should reconfigure it instead.

Lukas113 (approved):
LGTM

Lukas113 (approved):
Approved since we agreed to remove pinocchio.

PR #521, opened 2023-10-19
Remove PINOCCHIO from dependencies

lmachadopolettivalle:
We have agreed to remove PINOCCHIO from the Karabo package, since it was causing issues with its OpenMPI dependency. We will still have the separate PINOCCHIO fork in i4Ds, but it will not be included as part of Karabo installation, and can be used instead as a separate package.


Lukas113 (approved):
Removal of Pinocchio as a Karabo dependency as discussed in the meeting at 19.10.2023. 

I think everything is covered with this.

Issue #421, opened 2023-05-26
Design a good 95%-OSKAR-compatible-API for RASCIL


sfiruch:
Design a good 95%-OSKAR-compatible-API for RASCIL

_Originally posted by @deiruch in https://github.com/i4Ds/Karabo-Pipeline/issues/415#issuecomment-1563999299_

lmachadopolettivalle:
Current design approach:

(In the following, "framework" refers to the backends, e.g. OSKAR and RASCIL.

- [ ] Every resource class will store a framework agnostic dataset. E.g. `Observation` will store a series of floating point values for the configurations, `SkyModel` will store a DataFrame with columns corresponding to sky source properties, etc.
- [ ] Every resource class has methods to convert its data into objects compatible with any given framework. E.g. convert the DataFrame from `SkyModel` into a OSKAR-specific `array`, or a RASCIL-specific list of `skycomponents`

The resource classes that need to be reviewed and (if needed) modified are:
  - [ ] Observation
  - [ ] Telescope
  - [ ] SkyModel
  - [ ] Interferometer
  - [ ] Visibility
  - [ ] Image

PR #522, opened 2023-11-02
Telescope Class: Accept OSKAR and RASCIL backends

lmachadopolettivalle:
Implement changes described in #421 for the `Telescope` Class.

* Introduce a backend class via the `SimulatorBackend` enum
* Accept RASCIL as a backend to obtain its Telescope objects
* Modify how users can obtain a `Telescope` instance. Now, users can call e.g. `Telescope.constructor("MeerKAT")` instead of the more hard-coded `Telescope.get_MEERKAT_Telescope()`
* Modify `plot_telescope` to take into account which backend is being used
* Modify tests to use the new `constructor` method


karabo/simulation/telescope.py
```diff
@@ -27,11 +33,42 @@
...
+OSKAR_TELESCOPE_TO_VERSIONS = {
+    "ACA": ACAVersions,
+    "ALMA": ALMAVersions,
+    "ATCA": ATCAVersions,
+    "CARMA": CARMAVersions,
+    "NGVLA": NGVLAVersions,
+    "PDBI": PDBIVersions,
+    "SMA": SMAVersions,
+    "VLA": VLAVersions,
+}
```
Lukas113, line 70: This dict results in type: `Dict[str, Any]` if not specified further. However, all values are of type `enum.Enum`. If you define it as `Dict[str, Type[enum.Enum]]` (which is true), it will change some things which I'll explain later in this file.

karabo/simulation/telescope.py
```diff
@@ -79,6 +116,111 @@ def __init__(
...
+            is_version_provided = version is not None
+
+            if is_version_provided != is_version_required:
+                raise ValueError(
+                    """If a version if required, make sure to provided a version value.
+                If a version is not required, please do not provide a value."""
+                )
+
+            if is_version_provided and is_version_required:
+                if version not in accepted_versions:  # type: ignore[operator]
```
Lukas113, line 171: This is not a safe if-statement. The reason is because `version` can be None. Maybe you thought that it is not because of the comparison `if is_version_provided != is_version_required` before. However, if a non-valid `name` is provided and `version` is None (which is valid according to the signature), it results in a `TypeError` because `None not in Type[Enum]` is not allowed.

karabo/simulation/telescope.py
```diff
@@ -79,6 +116,111 @@ def __init__(
...
+                    raise ValueError(
+                        f"""{version} is not valid for telescope {name}.
+                        List of valid versions: {accepted_versions}."""
+                    )
+
+                datapath = datapath.format(cast(enum.Enum, version).value)  # noqa
```
Lukas113, line 177: unnecessary cast if you declare `OSKAR_TELESCOPE_TO_VERSIONS` as described above.

karabo/simulation/telescope.py
```diff
@@ -245,112 +403,75 @@ def get_cartesian_position(self) -> NDArray[np.float_]:
 
     @classmethod
     def read_from_file(cls, path: str) -> Optional[Telescope]:
-        if path.endswith(".tm"):
-            logging.info("Supplied file is a .tm file. Read as OSKAR Telescope file.")
-            return cls.read_OSKAR_tm_file(path)
-        else:
-            return None
+        raise DeprecationWarning("Use Telescope.read_OSKAR_tm_file(path) instead.")
```
Lukas113, line 406: I like it.

karabo/simulation/telescope.py
```diff
@@ -27,11 +33,42 @@
...
+    "MKATPlus": "mkatplus.tm",
+    "NGVLA": "ngvla-{0}.tm",
+    "PDBI": "pdbi-{0}.tm",
+    "SKA1LOW": "ska1low.tm",
+    "SKA1MID": "ska1mid.tm",
+    "SMA": "sma.{0}.tm",
+    "VLA": "vla.{0}.tm",
+    "VLBA": "vlba.tm",
+    "WSRT": "WSRT.tm",
+}
```
Lukas113, line 60: There could be a world where you define the keys of this dict as `Literal` and then type the dict accordingly. The advantage of this is that you can type the input of `constructor.name` with the literal which may be helpful. However, I still like you solution to check the value in `constructor` and raise an error if it's not correct there.

In my opinion it's worth to change that. However, if you have a reason not to do so, you can explain and then keep it as it is.

karabo/simulation/telescope.py
Lukas113: In case you want to be more specific, you can also define a `TypedDict` as follows:

from typing import TypedDict, Type

class OskarTelescopeVersionsType(TypedDict):
    ACA: Type[ACAVersions]
    ALMA: Type[ALMAVersions]
    ATCA: Type[ATCAVersions]
    CARMA: Type[CARMAVersions]
    NGVLA: Type[NGVLAVersions]
    PDBI: Type[PDBIVersions]
    SMA: Type[SMAVersions]
    VLA: Type[VLAVersions]

OSKAR_TELESCOPE_TO_VERSIONS: OskarTelescopeVersionsType = {
    "ACA": ACAVersions,
    "ALMA": ALMAVersions,
    "ATCA": ATCAVersions,
    "CARMA": CARMAVersions,
    "NGVLA": NGVLAVersions,
    "PDBI": PDBIVersions,
    "SMA": SMAVersions,
    "VLA": VLAVersions,
}

The advantage is, that you have the true type as `reveal_type` for the type-checker & the user by accessing the dict. So it's more save to use it. On the other hand, a disadvantage is that it seems a bit redundant.

karabo/simulation/telescope.py
Lukas113: In case the keys end up as Literal, you may also think about the keys of `OSKAR_TELESCOPE_TO_VERSIONS` to also be of the same or similar literal-type. At least to me it seems like the keys of the mentioned dict is a subset of `OSKAR_TELESCOPE_TO_FILENAMES`.

You could end up with something like the following (not sure though if it's the optimal solution):

```
from typing import Literal, Union

OSKAR_TELESCOPE_TO_VERSIONS
OskarTelescopeVersionsKeyType = Literal[
    "ACA",
    "ALMA",
    "ATCA",
    "CARMA",
    "NGVLA",
    "PDBI",
    "SMA",
    "VLA",
]
OskarTelescopeFnamesKeyType = Union[OskarTelescopeVersionsKeyType, Literal[
        "EXAMPLE",
        "MeerKAT",
        "ASKAP",
        "LOFAR",
        "MKATPlus",
        "PDBI",
        "SKA1LOW",
        "SKA1MID",
        "VLBA",
        "WSRT",
    ]
]
```

karabo/simulation/telescope.py
lmachadopolettivalle: But in the case you described, `version` would be `None`, and we wouldn't reach the nested `if` (line 171) since the first `if` (line 170) requires `version` to be not `None`, right? Let me know if I missed something.

Lukas113: Yeah you're right. Maybe it has something to do how the dict was typed. I don't know.

However, at the current commit, the `# type: ignore[operator]` seems unnecessary since `reveal_type(version)` is not None anymore after the if-statement `if is_version_provided and is_version_required:`.

Please check it yourself. I think flake8 or some other tool should complain?

Lukas113:
Another note:

You have some Python3.10 type-annotation features in it. To do a Union with `|` is just allowed from Python3.10, and we're operating on Python3.9 at the moment. The reason why Pylance is not complaining is because there is a `from __future__ import annotations` at the start of the file. However, it's still not a valid operator in Python3.9. That's the reason why `mypy` complains in the tests.

karabo/simulation/telescope.py
Lukas113: nvm, my IDE fooled me.

karabo/simulation/telescope.py
```diff
@@ -79,6 +142,109 @@ def __init__(
...
+        if self.backend is SimulatorBackend.OSKAR:
+            return self.path
+        if self.backend is SimulatorBackend.RASCIL:
+            return self.RASCIL_configuration
+
+        raise ValueError(
+            f"""Unexpected: current backend is set to {self.backend},
+        but expected one of {SimulatorBackend}.
+        Verify the construction of this Telescope instance."""
+        )
```
Lukas113, line 246: the return type must be Python3.9 conform, meaning Union instead of `|`. We have an issue about upgrading to Python 3.10. However since it's not done yet, we have to use python-featrues <= 3.9

Lukas113 (changes requested):
LGTM, like the changes.

karabo/simulation/telescope.py
```diff
@@ -79,6 +142,109 @@ def __init__(
...
         self.stations: List[Station] = []
 
+        self.backend: SimulatorBackend = SimulatorBackend.OSKAR
+
+        self.RASCIL_configuration: Optional[Configuration] = None
+
+    @classmethod
+    def constructor(
+        cls,
+        name: OSKARTelescopesWithVersionType | OSKARTelescopesWithoutVersionType,
```
Lukas113, line 152: Python3.10 Union

Actually, I was surprised that it worked with `from __future__ import annotations` and it didn't throw a TypeError at runtime. However, for the sake of consistency of the repository, I think it's still better to make a regular Python3.9 Union.

Lukas113 (changes requested):
LGTM

Just one minor change-request for the sake of consistency.

Lukas113 (approved):
LGTM

PR #526, opened 2023-11-20
512 sarus

### LINKED ISSUE #189: Unit Tests for User Docker Images
We should have tests to detect when public docker images stop working. At least basic functionality should work.

### LINKED ISSUE #449: Make pybdsf work with numpy >1.22
HVOX needs Numpy >=1.23 (Mentioned by @Lukas113 in #426), our current feedstock is restricted to numpy=1.22 due to restrictions in pybdsf (https://github.com/lofar-astron/PyBDSF). 

In
https://github.com/lofar-astron/PyBDSF/blob/master/pyproject.toml
are the following dependencies:
- oldest-supported-numpy
- setuptools<65.6

This blocks newer numpy versions. We need to make this work in order to add new dependencies such as HVOX.

Lukas113:
closes #449 
closed #189 

For testing-purpose, this PR needs it's dev-dependencies, referenced in `environment.yaml` and `meta.yaml`. However, BEFORE merging into the default brach, they need to be adapted and the according builds of the [Karabo-Feedstock](https://github.com/i4Ds/Karabo-Feedstock/tree/main) must be built at more or less the same time.

This PR was initially intended to be much smaller. However, because I wasn't able to test my changes with the old setup, I also had to define a dev & test-setup for Karabo to not interfere with the live conda-wheel and docker-images. 

This PR also includes changes in the Krabo-Feedstock, where a lot of dependencies were adjusted. The reason was originally because we had some open-mpi builds, which needed to be adapted. However, during the time I was working on this PR, we decided to remove the mpi-builds with the PR #521 from Karabo. This may seem like the work I was doing there is not necessary. However, I also needed to adapt the build-string setup to enable best-practice conda-builds, which depends usually on the python-version, package-hash and build-nr. Only then, the conda installer will be able to select the dependency wheels properly and can decide whether to download & install a new wheel or take it from the cache. The same goes for dev-builds. In addition, I loosened well-known and stable dependency-constraints. This had a positive side-effect that we are no more constrained as we used to be regarding upgrading dependencies, which solves the issue #449 .

This PR has the following changes in this repository:

- Intoducing dev- and test-setup to GitHub workflows
- Changed from open-mpi to mpich
- Added mpi4py tests
- Made codecov & pytest compatible with mpi-tests
- Loosened non-feedstock dependencies like `numpy`, `dask` or `xarray`.
- Included dev-builds of feedstock with different dependency-resolving to Karabo (need to be replaced before merging)
- Introducing dynamic versioning of Karabo using `versioneer`
- Changed packaging setup using `pyproject.toml` instead of `setup.cfg` -> got rid of requirements.txt
- Changed setup of user & dev- Dockerfile and made it theoretically ready for native mpich-hook (hook is disabled atm, because compilation takes too much time & memory) -> Details see Dockerfiles
- Adapted docs accordingly
- bugfix filter-sources for `xarray>2023.2`
- bugfix tests with cuda if cuda-is-available, but cuda-tests are disabled
- updated deprecated & buggy code in Karabo & it's dependencies for newer package-versions like `numpy` or `xarray`


Lukas113:
There are still issues:
- Testing bdsf can fail because: `ImportError: libboost_python39.so.1.82.0: cannot open shared object file: No such file or directory`. Not sure if bdsf is poining to an exact libbost-file since it's compiled. If yes, then we need to resolve & find a with the Karabo dependencies resolvable libboost-version first, and then fix it in the build, so that the exact version gets installed in the karabo-env again.
- Installation of Karabo can fail, because `karabo-pipeline-0.21.dev0-py39_1000 requires aratmospy 1.0.dev0.*, but none of the providers can be installed`. Have to build Karabo-dev probably new with adapted feedstock-builds.
- Docker-image with the wrong tag & labels got build if they were triggered from `conda-build` workflow. Probably because the arguments are not passed through correctly or there is an error in an if-statement?

Lukas113:
Ok, regarding `boost` I found out what the issue is/was. The issue is, that  boost, always need to be pinned at runtime to the version that was present at build time. E.g. `casacore` have different build-numbers of the same package-version for each supported boost-version. Fortunately, [conda build](https://docs.conda.io/projects/conda-build/en/stable/resources/variants.html#pinning-at-the-variant-level) provides a solution to this issue by using `pin_run_as_build`, where you do a build for each `boost` version, and pin it to each build.

Lukas113:
There's currently an issue with the GitHub runners (see [issue 3025](https://github.com/actions/runner/issues/3025)) which prevents me from properly testing the workflows before they live in the default-branch.

kenfus:
For me this is pretty hard to review. Anything I should focus on? Overall, it seems OK but like I said, it's hard to test or review.

Lukas113:
@kenfus  Just do a review wherever you feel you can provide some feedback. If you don't feel comfortable about reviewing on details, I suggest just focus on the design-choices.

Lukas113:
Ok, now merging should be fine. I've fixed feedstock-dependencies of karabo and in the feedstock itself to a fixed build-number. This should ensure, that future updates of the same package-version with a higher build-number don't affect older installations. I've tested the installation locally, and everything seems to work fine:

image

I hope that I didn't break older conda-wheels because of the updates I did. However, I don't see a good way how to fix that in case they would be broken. However, with the new setup of fixed build-numbers, this shouldn't happen anymore in the future.

The current implementation of fixed build-numbers in the feedstock is currently a little bit messy, because they're hard-coded there. However, I think the feedstock needs a refactoring anyway. But until we get to this point, I think we can leave it as it is at the moment.

As mentioned, the current Dockerfiles don't allow a native mpi-hook by sarus, because the mpi-implementation doesn't live in a standard location. @deiruch mentioned, that mpich should be installable via yum or apt. Maybe this makes the installation of mpi more easy and could be done on the github-runners. However, I think this should be done in a separate PR since this PR is already too big which is totally my fault.

karabo/imaging/image.py
```diff
@@ -501,7 +501,8 @@ def plot_power_spectrum(
         plt.gca().set_ylabel("Brightness temperature [K]")
         plt.gca().set_xscale("log")
         plt.gca().set_yscale("log")
-        plt.gca().set_ylim(1e-6 * np.max(profile), 2.0 * np.max(profile))
+        max_profile = float(np.max(profile))
+        plt.gca().set_ylim(1e-6 * max_profile, 2.0 * max_profile)
```
deiruch, line 505: Does this belong in this PR? Don't quite understand what this is about.

deiruch:
LGTM, apart from the mpi-hook/mpich-compilation things we talked about last week. (Use binary mpich release, instead of recompiling it)

karabo/imaging/image.py
Lukas113: If I remember correctly this was about mypy complaining about something. I used numpy 1.26 instead of 1.22

Lukas113:
Ok, I'll have a look if it works locally and on the runners.

Lukas113:
@deiruch 
mpich is installable through `apt`. However, the only version available for ubuntu-22.04 is 4.0-2. So to ensure compatibility, I tried to fix mpich to 4.0 in the environment.yaml. However, this is not a resolvable environment, which got the error-mesage posted below. The issue is when `h5py` is fixed to any mpich-build, the environment is not resolvable. Thus, the only way I see to resolve this is to make a decision.

Either, don't allow a native mpi-hook, but instead get a `h5py` mpich-wheel
OR downgrade `mpich` from 4.1 to 4.0 and dissalow an mpich-wheel of `h5py`

Since I don't know the consequences of either decision, I'm not sure which one is better or worse.

```
(karabo) root@9616a2df8d03:/repo# conda env update -f="environment.yaml"
Channels:
 - i4ds
 - nvidia/label/cuda-11.7.0
 - conda-forge
 - defaults
Platform: linux-64
Collecting package metadata (repodata.json): done
Solving environment: failed
Channels:
 - i4ds
 - nvidia/label/cuda-11.7.0
 - conda-forge
 - defaults
Platform: linux-64
Collecting package metadata (repodata.json): done
Solving environment: failed

LibMambaUnsatisfiableError: Encountered problems while solving:
  - package h5py-3.4.0-mpi_mpich_py310h83ac6b3_2 requires python >=3.10,<3.11.0a0, but none of the providers can be installed

Could not solve for environment specs
The following packages are incompatible
├─ bluebild 0.1.0 *_0 is installable with the potential options
│  ├─ bluebild 0.1.0 would require
│  │  └─ python >=3.9,<3.10.0a0  with the potential options
│  │     ├─ python [3.9.0|3.9.1|...|3.9.7], which can be installed;
│  │     ├─ python [3.9.0|3.9.1|...|3.9.9] would require
│  │     │  └─ python_abi 3.9.* *_cp39, which can be installed;
│  │     └─ python [3.9.10|3.9.12|3.9.16|3.9.17|3.9.18] would require
│  │        └─ python_abi 3.9 *_pypy39_pp73, which can be installed;
│  └─ bluebild 0.1.0 would require
│     ├─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
│     └─ python_abi 3.10.* *_cp310, which conflicts with any installable versions previously reported;
├─ h5py * mpi_mpich* is installable with the potential options
│  ├─ h5py [2.10.0|2.9.0|...|3.6.0], which can be installed;
│  ├─ h5py [2.10.0|3.1.0] would require
│  │  └─ python >=3.6,<3.7.0a0 , which conflicts with any installable versions previously reported;
│  ├─ h5py [2.10.0|3.1.0|...|3.6.0] would require
│  │  └─ python >=3.7,<3.8.0a0 , which conflicts with any installable versions previously reported;
│  ├─ h5py [2.10.0|3.1.0|...|3.6.0] would require
│  │  └─ python >=3.8,<3.9.0a0 , which conflicts with any installable versions previously reported;
│  ├─ h5py [3.4.0|3.6.0] would require
│  │  └─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
│  ├─ h5py [3.7.0|3.8.0] would require
│  │  ├─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
│  │  └─ python_abi 3.10.* *_cp310, which conflicts with any installable versions previously reported;
│  ├─ h5py [3.7.0|3.8.0] would require
│  │  ├─ python >=3.11,<3.12.0a0 , which conflicts with any installable versions previously reported;
│  │  └─ python_abi 3.11.* *_cp311, which conflicts with any installable versions previously reported;
│  ├─ h5py 3.7.0 would require
│  │  ├─ python >=3.7,<3.8.0a0 , which conflicts with any installable versions previously reported;
│  │  └─ python_abi 3.7.* *_cp37m, which conflicts with any installable versions previously reported;
│  ├─ h5py [3.7.0|3.8.0] would require
│  │  ├─ python >=3.8,<3.9.0a0 , which conflicts with any installable versions previously reported;
│  │  └─ python_abi 3.8.* *_cp38, which conflicts with any installable versions previously reported;
│  ├─ h5py [3.7.0|3.8.0] would require
│  │  └─ hdf5 >=1.12.2,<1.12.3.0a0 mpi_mpich_* with the potential options
│  │     ├─ hdf5 1.12.2 would require
│  │     │  └─ openssl >=1.1.1q,<1.1.2a , which can be installed;
│  │     ├─ hdf5 1.12.2, which can be installed;
│  │     └─ hdf5 [1.12.2|1.14.0] would require
│  │        └─ openssl >=1.1.1s,<1.1.2a , which can be installed;
│  ├─ h5py 3.7.0 would require
│  │  └─ hdf5 >=1.12.1,<1.12.2.0a0 mpi_mpich_* with the potential options
│  │     ├─ hdf5 1.12.1 would require
│  │     │  └─ openssl >=1.1.1l,<1.1.2a , which can be installed;
│  │     ├─ hdf5 1.12.1, which can be installed;
│  │     └─ hdf5 1.12.1 would require
│  │        └─ openssl >=1.1.1k,<1.1.2a , which can be installed;
│  ├─ h5py [3.10.0|3.9.0] would require
│  │  └─ mpich >=4.1.2,<5.0a0 , which can be installed;
│  ├─ h5py [3.10.0|3.9.0] would require
│  │  └─ python_abi 3.12.* *_cp312, which conflicts with any installable versions previously reported;
│  ├─ h5py 3.8.0 would require
│  │  └─ hdf5 >=1.14.0,<1.14.1.0a0 mpi_mpich_* with the potential options
│  │     ├─ hdf5 [1.12.2|1.14.0], which can be installed (as previously explained);
│  │     ├─ hdf5 1.14.0 would require
│  │     │  └─ openssl >=1.1.1t,<1.1.2a , which can be installed;
│  │     └─ hdf5 1.14.0, which can be installed;
│  ├─ h5py 3.9.0 would require
│  │  └─ mpich >=4.1.1,<5.0a0 , which can be installed;
│  └─ h5py 3.9.0 would require
│     └─ hdf5 >=1.14.1,<1.14.2.0a0 mpi_mpich_*, which can be installed;
├─ montagepy 6.0.0 *_0 is uninstallable because there are no viable options
│  ├─ montagepy 6.0.0 would require
│  │  ├─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
│  │  └─ python_abi 3.10.* *_cp310, which conflicts with any installable versions previously reported;
│  └─ montagepy 6.0.0 would require
│     ├─ python >=3.9,<3.10.0a0  with the potential options
│     │  ├─ python [3.9.0|3.9.1|...|3.9.7], which can be installed;
│     │  ├─ python [3.9.0|3.9.1|...|3.9.9], which cannot be installed (as previously explained);
│     │  └─ python [3.9.10|3.9.12|3.9.16|3.9.17|3.9.18], which can be installed (as previously explained);
│     └─ python_abi 3.9.* *_cp39 but there are no viable options
│        ├─ python_abi 3.9 conflicts with any installable versions previously reported;
│        └─ python_abi 3.9 would require
│           └─ python 3.9.* *_cpython, which conflicts with any installable versions previously reported;
├─ mpich 4.0**  is uninstallable because it conflicts with any installable versions previously reported;
├─ oskarpy 2.8.3 *_0 is uninstallable because there are no viable options
│  ├─ oskarpy 2.8.3 would require
│  │  ├─ oskar 2.8.3 *_0, which requires
│  │  │  └─ hdf5 >=1.14.2,<1.14.3.0a0  but there are no viable options
│  │  │     ├─ hdf5 1.14.2 conflicts with any installable versions previously reported;
│  │  │     └─ hdf5 1.14.2 would require
│  │  │        └─ openssl >=3.1.2,<4.0a0 , which conflicts with any installable versions previously reported;
│  │  └─ python_abi 3.10.* *_cp310, which conflicts with any installable versions previously reported;
│  └─ oskarpy 2.8.3 would require
│     ├─ oskar 2.8.3 *_0, which cannot be installed (as previously explained);
│     └─ python >=3.9,<3.10.0a0  with the potential options
│        ├─ python [3.9.0|3.9.1|...|3.9.7], which can be installed;
│        ├─ python [3.9.0|3.9.1|...|3.9.9], which cannot be installed (as previously explained);
│        └─ python [3.9.10|3.9.12|3.9.16|3.9.17|3.9.18], which can be installed (as previously explained);
├─ python 3.9**  is installable with the potential options
│  ├─ python [3.9.0|3.9.1|...|3.9.7], which can be installed;
│  ├─ python [3.9.0|3.9.1|...|3.9.9], which cannot be installed (as previously explained);
│  └─ python [3.9.10|3.9.12|3.9.16|3.9.17|3.9.18], which can be installed (as previously explained);
├─ ska-gridder-nifty-cuda 0.3.0 *_0 is uninstallable because there are no viable options
│  ├─ ska-gridder-nifty-cuda 0.3.0 would require
│  │  └─ python_abi 3.9.* *_cp39 but there are no viable options
│  │     ├─ python_abi 3.9 conflicts with any installable versions previously reported;
│  │     └─ python_abi 3.9, which cannot be installed (as previously explained);
│  └─ ska-gridder-nifty-cuda 0.3.0 would require
│     └─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
├─ ska-sdp-datamodels 0.1.3 *_0 is installable and it requires
│  └─ h5py >=3.7.0  with the potential options
│     ├─ h5py [3.7.0|3.8.0], which cannot be installed (as previously explained);
│     ├─ h5py [3.7.0|3.8.0], which cannot be installed (as previously explained);
│     ├─ h5py 3.7.0, which cannot be installed (as previously explained);
│     ├─ h5py [3.7.0|3.8.0], which cannot be installed (as previously explained);
│     ├─ h5py [3.7.0|3.8.0], which can be installed (as previously explained);
│     ├─ h5py 3.7.0, which can be installed (as previously explained);
│     ├─ h5py [3.10.0|3.9.0], which can be installed (as previously explained);
│     ├─ h5py [3.10.0|3.9.0], which cannot be installed (as previously explained);
│     ├─ h5py 3.8.0, which can be installed (as previously explained);
│     ├─ h5py 3.9.0, which can be installed (as previously explained);
│     ├─ h5py 3.9.0, which can be installed (as previously explained);
│     └─ h5py [3.10.0|3.7.0|3.8.0|3.9.0] conflicts with any installable versions previously reported;
└─ tools21cm 2.0.2 *_0 is installable and it requires
   └─ python 3.9.*  with the potential options
      ├─ python [3.9.0|3.9.1|...|3.9.7], which can be installed;
      ├─ python [3.9.0|3.9.1|...|3.9.9], which cannot be installed (as previously explained);
      └─ python [3.9.10|3.9.12|3.9.16|3.9.17|3.9.18], which can be installed (as previously explained).

```

deiruch:
For now we don't need h5py with mpi. It's only useful to allow parallel writing of HDF5 files, which we don't use AFAIK (see https://docs.h5py.org/en/stable/mpi.html).

Lukas113:
I've discovered a bug which I have to fix. Currently, for docker-containers everything works as intended. However, on Singularity and Sarus containers, the conda-setup doesn't work yet. This means that there, the wrong interpreter is selected per default. Also switching venvs is not straight-forward because `conda activate` doesn't work out of the bos.

For some reason, the `CMD` and `ENTRYPOINT` doesn't seem to work for non-docker containers (maybe just the way I used them, or entirely? not sure though). However, I need them because otherwise the `.bashrc` won't get executed, which is essential to switch a conda venv. Note: we need to have a venv-setup to not accidentally damage the base-environment. 
 
 Currently, I'm not sure how to solve this issue. Therefore, it's work in progress.

Lukas113:
There's also a bug with the mpich-hook. The installation in the container (conda dep resolving & apt-installation) and the conda-link to standard-location worked. However, the hook didn't. This is also work in progress.

Lukas113:
The issue of not being able to switch venvs in singularity containers is not straight-forward. The issue lies how docker and singularity handle user-ids and what directories and files get mounted by singularity. The issue regarding user-ids is, that one enters the a singularity container with the same user-id as the user who launched the process, while the files & directories of docker are created from root. A detailed description of the issue can be seen [here](https://geniac.readthedocs.io/en/latest/conda.html#how-can-a-conda-environment-be-activated-in-docker-or-singularity-containers). Currently I'm exploring solutions to this issue. I want to avoid creating a Singularity file, because that would imply that one has to get this file before creating a singularity container which is not optimal.

Lukas113:
Ok, the singularity issue seems to be fixed. The following docker-commands did the job:

```
# set bash-env accordingly for interactive and non-interactive shells for docker & singularity
RUN mkdir opt/etc && \
    echo "conda activate karabo" >> ~/.bashrc && \
    cat ~/.bashrc | sed -n '/conda initialize/,/conda activate/p' > /opt/etc/conda_init_script
ENV BASH_ENV=/opt/etc/conda_init_script
RUN echo "source $BASH_ENV" >> /etc/bash.bashrc && \
    echo "source $BASH_ENV" >> /etc/skel/.bashrc && \
    echo "source $BASH_ENV" >> /etc/profile
```

It configures the conda init & activate process for each use-case, ineractive and non-interacive bash-shell for docker & singularity.

Lukas113:
Currently not working are Singularity containers due to permission issues & native mpi-hook of Sarus containers. As discussed in the meeting at 21-12-23, these will be fixed in another PR.

Lukas113:
I've tested a simple mpi-example on cscs `srun -N2 -n2 -C gpu -A sk05 sarus run --mpi ghcr.io/i4ds/karabo-pipeline:0.21.dev1 python -c 'print("ABCDEFG")'` with and without the mpi-hook and it worked. The only thing left to do is that karabo also works for singularity containers, which is work in progress.

environment.yaml
```diff
@@ -2,37 +2,41 @@ channels:
...
+  - python                  =3.9
+  - aratmospy               =1.0.0=*_0
   - astropy
-  - bdsf
-  - bluebild
+  - bdsf                    =1.10.2=*_0
+  - bluebild                =0.1.0=*_0
   - cuda-cudart
-  - dask=2022.12.1
+  - dask                    >=2022.12.1
```
deiruch, line 12: Why this change?

karabo/test/test_utils.py
```diff
@@ -1,33 +1,35 @@
...
-def test_get_gpu_memory():
-    memory = get_gpu_memory()
-    assert isinstance(memory, int)
-    assert memory > 0
 
+CUDA_AVAILABLE = is_cuda_available()
 
-@pytest.mark.skipif(RUN_GPU_TESTS, reason="Does not fail when GPU is available")
+
+@pytest.mark.skipif(CUDA_AVAILABLE, reason="Doesn't make sense if cuda is available")
```
deiruch, line 15: Shouldn't this be `not CUDA_AVAILABLE` ?

karabo/util/file_handler.py
```diff
@@ -11,32 +13,82 @@
...
-        root_parent = os.getcwd()
-    root_dir = os.path.join(root_parent, karabo_folder)
-    return os.path.abspath(root_dir)
+def _get_tmp_dir() -> str:
+    """Gets the according tmpdir.
+
+    Defined env-var-dir > scratch-dir > tmp-dir
+
+    Honors TMPDIR and TMP environment variable(s).
+    The only thing not allowed is a collision between the mentioned env-vars.
```
deiruch, line 22: Unclear? What is a "collision"? Why?

karabo/util/file_handler.py
```diff
@@ -11,32 +13,82 @@
...
-    karabo_folder = "karabo_folder"
-    scratch = os.environ.get("SCRATCH")
-    if scratch is not None:
-        root_parent = scratch
-    else:
-        root_parent = os.getcwd()
-    root_dir = os.path.join(root_parent, karabo_folder)
-    return os.path.abspath(root_dir)
+def _get_tmp_dir() -> str:
+    """Gets the according tmpdir.
```
deiruch, line 17: "According" to what? (Perhaps "according" should be removed instead?) Unclear...

karabo/util/file_handler.py
```diff
@@ -11,32 +13,82 @@
...
+
+    Defined env-var-dir > scratch-dir > tmp-dir
```
deiruch, line 19: Unclear. Is this about priorities? Which directory is tried first?

karabo/util/file_handler.py
```diff
@@ -11,32 +13,82 @@
...
+                )
+        else:
+            tmpdir = os.path.abspath(TMP)
+            env_check = TMP
+            environment_varname = "TMP"
+    return tmpdir
+
+
+def _get_cache_dir() -> str:
+    """Gets a default cache-dir.
```
deiruch, line 54: This returns a new directory every time it's called, right? The name and description read as if you'd get the "default cache-dir" each and every time.  If the directory should be constant per session, we should store it somewhere, right?

deiruch (changes requested):
👍 Mostly minor details, but also some things I'd think we should double-check.

doc/src/development.md
```diff
@@ -28,31 +28,23 @@ Then create a local development environment with the provided `environment.yaml`
 conda env create -n <your-env-name> -f environment.yaml
 ```
 
-Then install the development dependencies using `requirements.txt`.
+Then install karabo as a package and the development dependencies.
```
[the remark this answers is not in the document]
Lukas113: I don't understand what the question is. Do you suggest I should correct karabo to Karabo with a capital K?

environment.yaml
Lukas113: I thought we discussed this matter? I'd like to not constrain well-maintained dependencies too much, so that Karabo is compatible with more other packages.

karabo/test/test_utils.py
Lukas113: This was a bugfix from my side. If you look at the test, `get_gpu_memory` will raise a RuntimeError in case there is no gpu. So if there is a gpu, this test will fail. Therefore there is no `not`.

karabo/util/file_handler.py
Lukas113: Hmm, I thought the terminology of "collision" is not ambiguous in this context. It means that the mentioned env-vars can't be different, because then the selection of a tmp-dir is ambiguous. Either both of them are the same of just one is set, which is represented in the code.

karabo/util/file_handler.py
Lukas113: According to what's written in the docstring.

Lukas113: But if it's confusing I can remove it. I just thought, than it's clear it's not a "a tmpir", because usually it's the same for each time you call the function.

karabo/util/file_handler.py
Lukas113: Yes it's about priorities, and the priority is depicted with greater than sign `>`. I agree that the docstring itself is not 100% self-explaining, but with the code it's meaning should be clear (I think). The reason why I didn't do the effort to describe it entirely is because it's a private function and therefore just for developers.

karabo/util/file_handler.py
Lukas113: I agree, default is not the best choice. However, `FileHandler` (and also this function) is refactored in #535 . I this PR, I just tried to solve the buggy tmpdir selection on a singularity-container & replaced the previous uuid-setup with 10 chars&digits which is what all tmpdir tools are doing as far as I've seen. But because this was not enough to make Singularity work, I did the rest in #535 .

    """Creates cache-dir-name.

    dir-name: karabo-<LTM|STM>-($USER-)<10-rnd-asci-letters-and-digits>

    The random-part of the cache-dir is seeded for relocation purpose.

    Returns:
        cache-dir-name
    """

karabo/__init__.py
```diff
@@ -23,12 +27,14 @@
     os.execv(sys.executable, ["python"] + sys.argv)
 
 # Setup dask for slurm
-from karabo.util.dask import prepare_slurm_nodes_for_dask
+if "SLURM_JOB_ID" in os.environ:
+    # ugly workaraound to not import stuff not available at build-time, but on import.
```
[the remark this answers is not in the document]
Lukas113: Usually, I try to keep code-comments as compact as possible, and therefore that's the description I came up with. I think the `, but on import` can be confusing. After reading the description, do you have a suggestion?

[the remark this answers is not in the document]
Lukas113: I removed the `, but on import` part

karabo/util/file_handler.py
Lukas113: I've rewritten the text & error message, so it's hopefully better explained.

karabo/test/test_utils.py
Lukas113: Only makes sense if cuda is not available

karabo/util/file_handler.py
Lukas113: write with Raises in docstring

PR #527, opened 2023-11-20
Use valid HI source catalogs, improve catalog access methods

lmachadopolettivalle:
- Remove SOUTH_CATALOG_FACTOR parameter, and instead rely on correct source catalogs
- Use updated HI source catalog with corrected 21cm flux values
- Deprecate old catalogs with incorrect flux values
- Move away from methods that are too specific, and rely on general file access methods instead


Lukas113 (approved):
LGTM

PR #540, opened 2024-01-29
535 singularity

### LINKED ISSUE #535: Karabo works with Singularity Container
Currently, Karabo doesn't work properly with Singularity containers. The issue is, that in a Singularity container, you're the same user which has launched a container, which is a different user than the one who created the image. This leads to read-only file-systems on each directory which is not mounted. This creates a situation, where you would have to tell a user, which is using Singularity, to use it in a very specific way, which is usually not what is desired.

Therefore, I suggest to unify each IO-operation with the filesystem using `FileHandler`. There you can ensure, that IO-operations should happen at a location, where the read-only constraint is not present.

In addition, I think it would also be a good idea to differentiate between long- and short-term cache, and implement a good caching strategy. I think this is important, to ensure that downloaded objects are accessible in different processes (e.g. mpirun) and runs (e.g. multiple jobs), so that you don't have to download it for each time & can download it into a writeable-location. At best, something that gets mounted, so you don't lose it after a container is destroyed.

Lukas113:
closes #535 

This PR addresses the issues of #535 . The main work was to make sure, that the disk-caching was not performed in a read-only filesystem, since in Singularity you're not the same user as the creator of the image. Therefore, I refactored `FileHandler` to have a short-term-memory and long-term-memory disk-cache available. Long-term is mainly for downloaded objects, where short-term-memory is for intermediate data-products, if no file/dir-path was provided in the according API-function. For details, see code & docstring of `FileHandler`.

In addition, I've seen two instances where I was not happy how it was implemented (in addition with the disk-caching), and saw a good opportunity to refactor the code. The first one was, how `DaskHandler` was implemented. It had some twisted code with SLURM, where the separation of concerns was not fulfilled. In addition, I've refactored some telescope-function code to be increase the stability & re-usability of the code.

This PR should be merged after #526 is merged, because it's a branch created from `512_sarus`.

The testing-setup was, I tested our tests in singularity- sarus & docker-container, and they passed there. This is of course no guarantee that I got everything, however, I've looked through the entire repository & the tests passed which is good enough I think.


doc/src/examples/example_structure.md
```diff
@@ -62,29 +62,24 @@ Leverage the `parallelize_with_dask` utility in Karabo to harness the power of p
...
+Karabo manages all available nodes through Dask, making the computational power conveniently accessible for the user. The `DaskHandler` class streamlines the creation of a Dask client and offers a user-friendly interface for interaction. This class contains static variables to midify the behavior of a Dask client, if they've changed before creating a client. 
 
-While users are not required to interact with Dask directly - thanks to the background processes managed by Karabo - the Dask client has to be initialized at the beginning of your script with `DaskHandler.setup` (see example below). This has to do with the spawning of new processes when creating `Nanny` processes.
+While users are not required to interact with Dask directly - thanks to the background processes managed by Karabo - the Dask client should to be initialized at the beginning of your script with `DaskHandler.setup` (see example below). This has to do with the spawning of new processes when creating `Nanny` processes.
```
lmachadopolettivalle, line 80: should to be -> should be

karabo/data/external_data.py
```diff
@@ -46,28 +42,44 @@ def download(url: str, local_file_path: FilePathType) -> int:
...
+        local_cache_dir = FileHandler().get_tmp_dir(
+            prefix="objects-download",
+            term="long",
+            purpose=purpose,
+        )
         local_file_path = os.path.join(
             local_cache_dir,
-            os.path.join(*remote_file_path.split("/")),  # convert to local filesys sep
+            os.path.join(
+                *remote_file_path.split(DownloadObject.split)
```
lmachadopolettivalle, line 74: Instead of manually handling the split of filepaths, we could use `pathlib` throughout this file: https://docs.python.org/3/library/pathlib.html#basic-use

It handles path splitting, using the appropriate separator based on the OS, and can also handle mkdir and rm calls.

karabo/simulation/interferometer.py
```diff
@@ -236,31 +235,30 @@ def __init__(
...
-        if use_dask is None and not client:
-            print(
-                KaraboWarning(
-                    "Parameter 'use_dask' is None! Using function "
-                    "'karabo.util.dask.DaskHandler.should_dask_be_used()' "
-                    "to overwrite parameter 'use_dask' to "
-                    f"{DaskHandler.should_dask_be_used()}."
+        self.use_gpus = use_gpus
+
+        if use_dask is True or client:
```
lmachadopolettivalle, line 246: In these ifs and elses, add parentheses to make the logic more explicit and readable. E.g. `if (use_dask is True) or (client is not None)`

karabo/simulation/interferometer.py
```diff
@@ -407,13 +405,14 @@ def __run_simulation_parallized_observation(
 
         # Scatter sky
         array_sky = self.client.scatter(array_sky)
-        fh = FileHandler.get_file_handler(
-            self, prefix=self._fh_prefix, verbose=self._fh_verbose
+        tmp_dir = FileHandler().get_tmp_dir(
+            prefix="simulation-praallezed-observation-",
```
lmachadopolettivalle, line 409: simulation-praallezed-observation -> simulation-parallelized-observation

karabo/simulation/telescope.py
```diff
@@ -615,53 +626,151 @@ def __float_try_parse(cls, value: str) -> float:
...
+        if (n_stations_layout := stations.shape[0]) != (n_stations := df_tel.shape[0]):
+            raise KaraboError(
+                f"Number of stations missmatch of {n_stations_layout=} & {n_stations=}"
+            )
+        df_tel["x"] = stations[:, 0]
+        df_tel["y"] = stations[:, 1]
+        return df_tel
+
+    @classmethod
+    def _get_number_str(cls, num: int, n_digits: int) -> str:
```
lmachadopolettivalle, line 675: No need for this method, we can use `str.zfill` instead (builtin): https://docs.python.org/3/library/stdtypes.html#str.zfill

Example:
```
>>> a = "1234"
>>> a.zfill(3)
'1234'
>>> a.zfill(9)
'000001234'
```

karabo/util/file_handler.py
```diff
@@ -81,158 +121,261 @@ class FileHandler:
...
-    def remove_empty_dirs(consider_fh_dir_identifier: bool = True) -> None:
-        """Removes emtpy directories in `FileHandler.root`.
+    def clean(
+        term: _LongShortTermType = "short",
+    ) -> None:
+        """Removes the entire directory specified by `term`.
 
-        Just manual use recommended since it doesn't consider directories which
-         are currently in use and therefore it could interrupt running code.
+        Be careful with cleaning, to not mess up dirs of other processes.
```
lmachadopolettivalle, line 304: Is this a suggestion to the user, or a description of what the method is doing ("being careful")? Change wording to make it more explicit. If it is a suggestion to the user, add a note on how they should be careful.

karabo/util/file_handler.py
```diff
@@ -52,25 +65,52 @@ def _get_disk_cache_root() -> str:
     return tmpdir
 
 
-def _get_cache_dir() -> str:
-    """Creates a deterministic & user-specific cache-dir-name.
+def _get_rnd_str(k: int, seed: _SeedType = None) -> str:
+    """Creates a random ascii+digits string with length=`k`.
+
+    Most tmp-file tools are using a sting-length of 10.
```
lmachadopolettivalle, line 71: sting -> string

karabo/util/file_handler.py
```diff
@@ -52,25 +65,52 @@ def _get_disk_cache_root() -> str:
...
+        seed: Seed.
+
+    Returns:
+        Random generated string.
+    """
+    random.seed(seed)
+    return "".join(random.choices(string.ascii_letters + string.digits, k=k))
+
+
+def _get_cache_dir(term: _LongShortTermType) -> str:
```
lmachadopolettivalle, line 84: Very nice

karabo/util/dask.py
```diff
@@ -1,494 +1,747 @@
...
-        a Dask client will be created.
-
-
-
+    use_dask:
+        Whether to use Dask or not. If None, then Karabo will decide whether
+        to use dask or not for certain tasks.
+    use_processes:
+        Use processes instead of threads?
+        Threads:
```
lmachadopolettivalle, line 47: Very helpful comments on processes vs threads 👍

karabo/util/dask.py
```diff
@@ -1,494 +1,747 @@
...
     """
 
     dask_client: Optional[Client] = None
-    n_workers_scheduler_node: int = 1
-    memory_limit: Optional[int] = None
+    memory_limit: Optional[float] = None
     n_threads_per_worker: Optional[int] = None
     use_dask: Optional[bool] = None
-    use_workers_or_nannies: Optional[str] = "nannies"
     use_proccesses: bool = False  # Some packages, such as pybdsf, do not work
```
lmachadopolettivalle, line 61: Keep the second line of this comment to make it more clear (i.e. bring back the part "# with processes because they spawn subprocesses.")

karabo/util/dask.py
```diff
@@ -1,494 +1,747 @@
...
+    @classmethod
+    def get_dask_client(cls) -> Client:
+        """Get (create if not exists) a dask-client.
+
+        Returns:
+            Dask-client.
+        """
+        if cls.dask_client is not None:
+            return cls.dask_client
+        if MPI.COMM_WORLD.Get_size() > 1:  # TODO: testing of whole if-block
```
lmachadopolettivalle, line 80: Is this TODO still needed?

karabo/util/dask.py
```diff
@@ -1,494 +1,747 @@
...
-                    DaskHandler.n_workers_scheduler_node, DaskHandler.memory_limit
-                )
-            else:
-                DaskHandler.dask_client = get_local_dask_client(
-                    DaskHandler.memory_limit
-                )
-            # Write the dashboard link to a file
-            with open("karabo-dask-dashboard.txt", "w") as f:
-                f.write(DaskHandler.dask_client.dashboard_link)
+            cls.dask_client = Client(processes=cls.use_proccesses)  # TODO: testing
```
lmachadopolettivalle, line 86: TODO here as well

lmachadopolettivalle (changes requested):
Very well-written code, only a few small comments about documentation and a few methods.

karabo/data/external_data.py
Lukas113: I think there's a misunderstanding here. This is just a conversion from url-separator to os-separator.

lmachadopolettivalle (approved):
💯

Issue #542, opened 2024-02-14
allow xarray > 2023.2


Lukas113:
For the imaging part of Karabo while running `source_detection.ipynb`, I've encountered the following error with `xarray` 2023.11 (see traceback). When I downgraded `xarray` to 2023.2 (release before blame commit to raise ValueError), there was no such error and everything worked fine. Because this error is caused through `RASCIL` and `ska_sdp_func_python` (see traceback), I think fixing this error is out of my control. If my assessment is correct, when should fix it as follows:

First, just constrain `xarray` to be <= 2023.2. However, this also constrains `pandas` to be < 2 (conda-resolver said so). And because these two libraries are widely used libraries, we should address this issue further. However, this might require some effort. Firstly, because `RASCIL` hasn't contrained `ska_sdp_func_python` & `ska_sdp_func` versions, I'm concerned that the latest release of `RASCIL` 1.1.0 (April 2023) might not be compatible with the (hopefully fixed) major release of [ska-sdp-func 1.0](https://github.com/ska-telescope/ska-sdp-func/tags) (November 2023). In addition, we also need to build the according conda-wheels after testing.

Long story short: As a hotfix, I'll constrain `xarray` <= 2023.2 (and accept the `pandas` downgrade for the moment). The best-case scenario would be that `RASCIL` soon does a new release, where we could fix the according `ska-sdp` deps. I don't think that building a dirty conda-wheel (10 months after the last release) of `RASCIL` is a good solution. If there's a better suggestion, I'd like to hear it.

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[14], line 6
      2 imager_askap.ingest_vis_nchan = 16
      4 # Try differnet algorithm
      5 # More sources
----> 6 deconvolved, restored, residual = imager_askap.imaging_rascil(
      7     clean_nmajor=1,
      8     clean_algorithm="mmclean",
      9     clean_scales=[10, 30, 60],
     10     clean_fractional_threshold=0.3,
     11     clean_threshold=0.12e-3,
     12     clean_nmoment=5,
     13     clean_psf_support=640,
     14     clean_restored_output="integrated",
     15     use_dask=True,
     16 )

[...]

PR #543, opened 2024-02-14
542 xarray

Lukas113:
See issue #542 

Additional bug fixing and minor improvements

Unfortunately, I created this branch from 512_sarus and not from main, if you wonder why there are so many commits.


lmachadopolettivalle (approved):
LGTM 💯

Lukas113:
@kenfus is it okay to remove `guess_beam_parameters` in this PR? I removed it because it never worked. If you have a good reason to keep it and ensure that it works, I'll happily add the function to the module again.

conda/meta.yaml
```diff
@@ -55,7 +55,7 @@ requirements:
     - ska-sdp-datamodels      =0.1.3=*_0
     - ska-sdp-func-python     =0.1.4=*_0
     - tools21cm               =2.0.2=*_0
-    - xarray                  >=2022.10.0
+    - xarray                  <=2023.2  # see issue #542
```
sfiruch, line 58: Please fix version instead.

environment.yaml
```diff
@@ -36,7 +36,7 @@ dependencies:  # package-version & build-number of Karabo-Feedstock deps should
   - ska-sdp-datamodels      =0.1.3=*_0
   - ska-sdp-func-python     =0.1.4=*_0
   - tools21cm               =2.0.2=*_0
-  - xarray                  >=2022.10.0
+  - xarray                  <=2023.2  # see issue #542
```
sfiruch, line 39: Please fix version instead.

Lukas113:
An update, with pandas >=2, the following FutureWarning occured:

```
/home/lukas/miniconda3/envs/karabo_dev_env/lib/python3.9/site-packages/ska_sdp_datamodels/visibility/vis_model.py:74: FutureWarning: the `pandas.MultiIndex` object(s) passed as 'baselines' coordinate(s) or data variable(s) will no longer be implicitly promoted and wrapped into multiple indexed coordinates in the future (i.e., one coordinate for each multi-index level + one dimension coordinate). If you want to keep this behavior, you need to first wrap it explicitly using `mindex_coords = xarray.Coordinates.from_pandas_multiindex(mindex_obj, 'dim')` and pass it as coordinates, e.g., `xarray.Dataset(coords=mindex_coords)`, `dataset.assign_coords(mindex_coords)` or `dataarray.assign_coords(mindex_coords)`.
  super().__init__(data_vars, coords=coords, attrs=attrs)
  ```

File: karabo/sourcedetection/result.py [hunk 3]
```diff
@@ -110,7 +110,7 @@ def detect_sources_in_image(
         cls: Type[_SourceDetectionResultType],
         image: Image,
         beam: Optional[BeamType] = None,
-        quiet: bool = False,
+        verbose: bool = False,
```
sfiruch, line 113: Did you intend to change the default value?

(Asking because `verbose` is the opposite of `quiet`)

sfiruch (changes requested):

karabo/sourcedetection/result.py
Lukas113: I changed from quiet to verbose to not have different keywords for the same thing, even they're inverted.

The change to be not verbose is intentional, but based on personal preference. I just don't like default verbose output unless the process always takes a large amount of time (there, a `tqdm` bar would be nice).

Lukas113:
I also removed the build-string fixations, because they prevent updates on dependency-constraints. Therefore, I introduces the following instructions to the README of the Feedstock & triggered the according builds there.

**Updating an already existing Wheel**: It may happen that an update of a dependency (direct or transversal) introduces a breaking change in a future release. Then, the current constraints of an existing conda-wheel would be outdated an will most likely result in a broken environment. To prevent this, updating the according meta.yaml and increase the build-nr is not enough, because the solver doesn't just consider the largest build-numbers of conda-wheels. So, to fix this for current and also older releases, you have to relabel oudated conda-wheels to `legacy`. Also do a comment in the meta.yaml, why you did constrain the dependency. This makes future builds where the constraints may have to be reconsidered way easier. It was already mentioned in this paragraph, but don't forget to increase the build-number.

sfiruch (dismissed):

kenfus:
> @kenfus is it okay to remove `guess_beam_parameters` in this PR? I removed it because it never worked. If you have a good reason to keep it and ensure that it works, I'll happily add the function to the module again.

I rather not. This was requested by @rohitcbscient and it also works in the current example of the `source_detection.ipynb`.

sfiruch (approved):

PR #544, opened 2024-02-22
SkyModel Class: Accept OSKAR and RASCIL backends

lmachadopolettivalle:
Implement changes described in https://github.com/i4Ds/Karabo-Pipeline/issues/421 for the SkyModel Class.

After a design discussion on 08.02.2024 with @Lukas113 and @kenfus we decided on the following approach:
- Continue storing source catalog data in a 2D `numpy` array, inspired by the OSKAR data format
- Introduce a method in `SkyModel` to export sources as a list of RASCIL `SkyComponents`, which can then be used to compute visibilities and later dirty images
- Create tests for the conversion to RASCIL format

This design choice allows users to benefit from existing code in Karabo to load, filter and visualize SkyModel instances, while also allowing for RASCIL-specific interferometry later in the pipeline.


karabo/simulation/sky_model.py
```diff
@@ -1529,3 +1537,135 @@ def sky_from_h5_with_redshift_filtered(
...
+            # 1. Remove sources that fall outside all desired frequency channels
+            # 2. Assign each source to the frequency channel closest
+            # to its corresponding redshift (using np.digitize)
+            # 3. For each source, create a SkyComponent,
+            # with flux array equal to 0 on all channels,
+            # except for its closest channel, where all its flux will belong
+            # This is equivalent to having the source's SED equal to a delta function
+            # at the frequency corresponding to its redshift,
+            # which is true for line emission catalogues.
+            min_redshift, max_redshift = convert_frequency_to_z(
```
Lukas113, line 1597: This is something which is not allowed by `mypy`. The reason is, you provide an NDArray as input, which is also the output. However, variadic type-hints are not supported by mypy and Python3.9, which is needed for shape-typing. Without knowledge of the shape of an NDArray, there is no way to tell if the return value is an NDArray (thus Iterable) or a scalar value.

What you do here is direct unpacking of the NDArray. However, from a static point of view, `convert_frequency_to_z` has no way to know what the return type of an unpacked NDArray is because of the reasons mentioned.

The best solution which comes to mind is to use `cast` to something like `Iterable[np.float_]`, because you know the input is 2-d, and the output is 2-d again.

karabo/simulation/sky_model.py
```diff
@@ -1529,3 +1537,135 @@ def sky_from_h5_with_redshift_filtered(
...
+                right=False,
+            )
+
+            # E.g. if channel starts are [1e8, 2e8],
+            # then a source at frequency 1.5e8 should fall into the 0th channel.
+            # However, digitize returns index 1 for such a source.
+            # Therefore, we subtract 1 from the return value of np.digitize
+            source_channel_indices -= 1
+
+            skycomponents = []
```
Lukas113, line 1635: it helps to declare the generic type of empty lists, because they have no generic type-infos from the constructor.

Lukas113 (changes requested):
LGTM from my current point of view. There are some changes, but mainly to improve static-type hints.

Lukas113 (approved):
LGTM

PR #555, opened 2024-03-18
fixed install commands doc for WSL2

### LINKED ISSUE #550: Improve WSL-doc
Currently, just following the dev-install instructions on WSL will fail with a weird error at `pip install -e ".[dev]", because `usr/lib/wsl/lib` is not available in the path. 

The `conda env config vars set LD_LIBRARY_PATH=/usr/lib/wsl/lib` on the venv solves this issue, which is already in our documentation for the user-installation at the end. This should be more visible in the user and dev-installation doc at the point which is relevant (maybe already at the start of the section?)

Lukas113:
closes #550 

Hotfix to link shared libraries like cuda in `/usr/lib/wsl/lib`. Properly linking using `ldconfig` doesn't work, because NVIDIA doesn't  install `libcuda.so.1` as a symlink (see https://forums.developer.nvidia.com/t/wsl2-libcuda-so-and-libcuda-so-1-should-be-symlink/236301)

### DIFF
--- a/doc/src/development.md
+++ b/doc/src/development.md
@@ -26,12 +26,17 @@ Then create a local development environment with the provided `environment.yaml`
 
 ```shell
 conda env create -n <your-env-name> -f environment.yaml
+conda activate <your-env-name>
+```
+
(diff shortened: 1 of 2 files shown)


sfiruch (approved):
LGTM

PR #560, opened 2024-03-22
547 hotfix mightee

### LINKED ISSUE #547: Fix MIGHTEE Sky
The current implementation of `SkyModel.get_MIGHTEE_Sky` has some major errors from our side, because we don't extract and convert the sources and their parameters correctly from the .fits file. Our `SkyModel` treats the sources as follows:
image

 We do the following not correctly:
- Assignment of Stokes-I flux, currently, "NU_EFF" is taken, which are actually frequencies in HZ and not peak-Flux in Jy/beam (stored in "S_PEAK")
- major & minor axis of `SkyModel` are in arcsec, however, if the unit is different, like it is for the MIGHTEE sky (in deg), we need to convert them.
- Frequencies are just taken from `SkyModel.get_sky_model_from_fits` as input parameters and not from the .fits file. But they're availabe at "NU_EFF" and we just ignore them. They should be passed to `reference_frequency`.

As far as I understand, these are the errors we do for the MIGHTEE sky. @rohitcbscient please correct me if I'm mistaken in my statements or referencing the correct header-names.

As a hotfix, we should just raise an error, until this is fixed. Otherwise, you just get a corrupt sky.

Lukas113:
closes #547 

This PR addresses the following:
- Reading correct cols from .fits files of MIGHTEE and GLEAM
- Implemented a correct and generalized version to read sky-sources from .fits file. The previous version had the following issues:
    - Didn't consider unit-scaling & checks
    - Didn't allow to read frequencies
    - Was not generalized
- Introduced helper-classes with tests to address these issues


mpluess (approved):
Looks good as far as I can tell

PR #563, opened 2024-03-27
561 download object

### LINKED ISSUE #561: Improve DownloadObject
Currently, `DownloadObject` doesn't seem good enough and safe to me what it's doing. I propose the following changes:

- allow `download` for `verify=False` if necessary, but set True as default
- Just cleanup entire directory in `download` if it didn't exist before. Otherwise, just the partial downloaded file. Currently, it could remove directories which have other content if the download gets interrupted.
- Add progress-bar option to `download`. This [stackoverflow-post](https://stackoverflow.com/questions/37573483/progress-bar-while-download-file-over-http-with-requests) may be helpful.
- Extend `get_object` with the mentioned options.
- Set `verify=False` in `is_url_available` and ignore `InsecureRequestWarning` from `urllib3` (just in this function)
- Add docstrings to class

Lukas113:
closes #561 

Addresses all points mentioned in the issue.

### DIFF
--- a/karabo/data/external_data.py
+++ b/karabo/data/external_data.py
@@ -44,6 +44,7 @@ class DownloadObject:
         url: str,
         local_file_path: FilePathType,
         verify: bool = True,
+        verbose: bool = True,
     ) -> int:
         download_dir = os.path.dirname(local_file_path)
         dir_existed = False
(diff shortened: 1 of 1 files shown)


mpluess (approved):
LGTM

PR #564, opened 2024-03-27
splitted ltm- and stm-memory roots for FileHandler to avoid tmp-clear…

Lukas113:
Split long-term-memory disk-cache root from short-term-memory. The reason for this change is because before a differentiation between the two use-cases regarding the root-directory was not possible. Therefore, the default root-dir /tmp caused to remove long-term-memory files at each restart, which is not desired.

- The previous root is now the default for short-term-memory
- Long-term-memory directory now follows the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).


mpluess (approved):
LGTM

PR #565, opened 2024-04-02
Fix version

Lukas113:
This PR is fixing the issue that versioneer can't properly get the version from a package-wheel. It works for pip install from local, but not from conda install from anaconda.org. However, I've tested the changed by creating a conda dev-wheel and checked that `karabo.__version__ != "0+unknown"`

Current release installation:
image

dev-installation with these changes:
image


mpluess (approved):
LGTM

PR #566, opened 2024-04-05
562 mals

### LINKED ISSUE #562: Add MALS sky
Yesterday I got an e-mail that a new [MALS-survey](https://mals.iucaa.in/) from MeerKAT got released. Since MeerKAT is a precursor of SKA. I was wondering whether we should provide the survey through Karabo. It has a catalogue available for more than 700k unique sources, created using PyBDSF. @lmachadopolettivalle and @sfiruch what do you think about this?

Website: https://mals.iucaa.in/

CSV-col-desc: https://iopscience.iop.org/article/10.3847/1538-4365/acf7b9#apjsacf7b9t3

There are some inconveniences to provide the survey as a Karabo-Sky. One of them is that the full catalogue is just available as .csv and is larger than 3.6GB, which currently takes more than an hour to download from their server. However, I was able to download the .csv and convert it to a .fits file, where the remaining data is saved more efficiently and filtered with the relevant SkyModel.sources data. The code looks roughly as follows, where some parts obviously need some adaptions:

```python
from __future__ import annotations

import os
import shutil
import tempfile
import warnings
from collections.abc import Iterable, Iterator

import astropy.units as u
import numpy as np
import pandas as pd
import requests
from astropy.coordinates import Angle
from astropy.io import ascii, fits
from astropy.table.table import Table
from astropy.units import Quantity, UnitBase
from karabo.data.external_data import DownloadObject
from karabo.util._types import DirPathType, FilePathType


def create_MALS_survey_as_fits(
    directory: DirPathType,
    version: int = 3,
    check_for_updates: bool = False,
    verbose: bool = True,
) -> str:
    directory = str(directory)
    try:
        version = int(version)  # run-time check
    except ValueError as e:
        raise TypeError(f"{version=} must be an integer!") from e
    base_url = "https://mals.iucaa.in"
    remote_path = "/catalogue/"
    fname_csv_template = "catalogue_malsdr1v{0}_all.csv"
    fname_csv = fname_csv_template.format(version)
    url = base_url + remote_path + fname_csv
    if check_for_updates:
        update_address = base_url + remote_path + fname_csv_template.format(version + 1)
        if DownloadObject.is_url_available(update_address):
            print(
                f"An updated version of {url} is available! Please have a look at {base_url}"
            )
    local_csv_file = os.path.join(directory, fname_csv)
    if not local_csv_file.endswith(
        ".csv"
    ):  # should be impossible, but just for double-checking
        raise ValueError(f"{local_csv_file=} is not a valid csv file!")
    local_fits_file = local_csv_file[:-4] + ".fits.gz"
    if os.path.exists(local_fits_file):
        if verbose:
            print(f"{local_fits_file} already exists.")
        return local_fits_file
    if not os.path.exists(local_csv_file):
        _ = DownloadObject.download(
            url=url, local_file_path=local_csv_file, verify=False, verbose=verbose
        )  # about 3.6GB, may take VERY long

    if verbose:
        "Read csv from disk ..."
    df = pd.read_csv(local_csv_file)  # takes about 40-45 s
    cols_of_interest = [
        "ra_max",
        "dec_max",
        "peak_flux",
        "ref_freq",
        "maj",
        "min",
        "pa",
        "source_name",
        "pointing_name",  # uniqueness is only ensured for a combination of pointing-id and spw-id
        "spw_id",
    ]
    units: dict[str, UnitBase] = {
        "ra_max": u.deg,
        "dec_max": u.deg,
        "peak_flux": u.mJy / u.beam,
        "ref_freq": u.MHz,
        "maj": u.arcsec,
        "min": u.arcsec,
        "pa": u.deg,
    }
    table = Table.from_pandas(df[cols_of_interest], units=units)
    if verbose:
        print(f"Creating {local_fits_file} ...")
    with tempfile.TemporaryDirectory() as tmpdir:
        tmp_fits = os.path.join(tmpdir, "mals.fits")
        table.write(tmp_fits, format="fits")  # about 368MB
        with fits.open(tmp_fits) as f:
            f.writeto(local_fits_file)

    return local_fits_file


fits_file = create_MALS_survey_as_fits(
    directory="./",
)
```

```python
from karabo.simulation.sky_model import SkySourcesUnits, SkyModel, SkyPrefixMapping

fits_file = "catalogue_malsdr1v3_all.fits.gz"
unit_mapping: dict[str, UnitBase] = {
    "deg": u.deg,
    "beam-1 mJy": u.mJy / u.beam,
    "MHz": u.MHz,
    "arcsec": u.arcsec,
}
prefix_mapping = SkyPrefixMapping(
    ra="ra_max",
    dec="dec_max",
    stokes_i="peak_flux",
    ref_freq="ref_freq",
    major="maj",
    minor="min",
    pa="pa",
    id="source_name",
)
unit_sources = SkySourcesUnits(
    stokes_i=u.Jy / u.beam,
)
sky = SkyModel.get_sky_model_from_fits(
    fits_file=fits_file,
    prefix_mapping=prefix_mapping,
    unit_mapping=unit_mapping,
    units_sources=unit_sources,
    min_freq=976.4e6,
    max_freq=1036.5e6,
)
```

Lukas113:
closes #562 

- Adds MALS survey to Karabo
- Adds `SkySourcesUnits.get_pos_ids_to_ra_dec` to convert J2000 sky-source-ids to RA-DEC in degrees

### DIFF
--- a/karabo/simulation/sky_model.py
+++ b/karabo/simulation/sky_model.py
@@ -48,6 +48,7 @@ from xarray.core.coordinates import DataArrayCoordinates
 from karabo.data.external_data import (
     GLEAMSurveyDownloadObject,
     HISourcesSmallCatalogDownloadObject,
+    MALSSurveyV3DownloadObject,
     MIGHTEESurveyDownloadObject,
 )
 from karabo.error import KaraboSkyModelError
@@ -2080,6 +2081,66 @@ class SkyModel:
             max_freq=max_freq,
             encoded_freq=None,
             memmap=False,
+        )
+
+    @classmethod
+    def get_MALS_DR1V3_Sky(
+        cls: Type[_TSkyModel],
+        min_freq: Optional[float] = None,
+        max_freq: Optional[float] = None,
+    ) -> _TSkyModel:
+        """Creates a SkyModel containing "MALS V3" catalogue, of 715,760 sources,
+        where the catalogue and it's information are from 'https://mals.iucaa.in/'.
+
+        The MeerKAT Absorption Line Survey (MALS) consists of 1655 hrs of MeerKAT time
+        (anticipated raw data ~ 1.7 PB) to carry out the most sensitive search of HI
+        and OH absorption lines at 0<z<2, the redshift range over which most of the
+        cosmic evolution in the star formation rate density takes place. The MALS
+        survey is described in 'Gupta et al. (2016)'.
+
+        MALS sky-coverage consists of 392 sources trackings between all RA-angles
+        and DEC[-78.80,32.35].
+
+        MALS's frequency-range: 902-1644 MHz.
+
+        For puplications, please honor their work by citing them as follows:
+        - If you describe MALS or associated science, please cite 'Gupta et al. 2016'.
+        - If you use DR1 data products, please cite 'Deka et al. 2024'.
+
+        Args:
+            min_freq: Set min-frequency in Hz for pre-filtering.
+            max_freq: Set max-frequency in Hz for pre-filtering.
+
+        Returns:
+            MALS sky as `SkyModel`.
+        """
+        survey_file = MALSSurveyV3DownloadObject().get()
+        unit_mapping: dict[str, UnitBase] = {
+            "deg": u.deg,
+            "beam-1 mJy": u.mJy / u.beam,
+            "MHz": u.MHz,
+            "arcsec": u.arcsec,
+        }
+        prefix_mapping = SkyPrefixMapping(
+            ra="ra_max",
+            dec="dec_max",
+            stokes_i="peak_flux",
+            ref_freq="ref_freq",
+            major="maj",
+            minor="min",
+            pa="pa",
+            id="source_name",
+        )
+        unit_sources = SkySourcesUnits(
+            stokes_i=u.Jy / u.beam,
+        )
+        return cls.get_sky_model_from_fits(
+            fits_file=survey_file,
+            prefix_mapping=prefix_mapping,
+            unit_mapping=unit_mapping,
+            units_sources=unit_sources,
+            min_freq=min_freq,
+            max_freq=max_freq,
         )
 
     @classmethod
--- a/karabo/simulation/sky_model.py
+++ b/karabo/simulation/sky_model.py
@@ -4,7 +4,7 @@ import copy
 import enum
 import math
 import re
-from collections.abc import Hashable, Mapping
+from collections.abc import Hashable, Mapping, Sequence
 from copy import deepcopy
 from dataclasses import dataclass, fields
 from typing import (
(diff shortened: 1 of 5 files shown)


Lukas113:
@lmachadopolettivalle do you know if the J2000 source ids have a proper specification? If yes, I can probably improve the parsing a lot. I haven't found anything on the web.

karabo/simulation/sky_model.py
```diff
@@ -378,6 +379,114 @@ def extract_names_and_freqs(
                 names_and_freqs[col_name] = extracted_freq
         return names_and_freqs
 
+    @classmethod
+    def get_pos_ids_to_ra_dec(
```
lmachadopolettivalle, line 383: Nice method! I wonder if it would belong in SkyModel or if it'd be better to keep it in a separate module for helper functions. This is a small detail so feel free to keep it here if preferred.

karabo/test/test_skymodel.py
```diff
@@ -58,6 +60,10 @@ def test_mightee():
     _ = SkyModel.get_MIGHTEE_Sky(min_freq=1.3e9, max_freq=1.4e9)
 
 
+def test_mals_v3():
```
lmachadopolettivalle, line 63: We could assert something about the sources in this sky, e.g. that the sources data is not empty

lmachadopolettivalle (approved):
LGTM, left a few optional suggestions but feel free to merge as is

lmachadopolettivalle:
> @lmachadopolettivalle do you know if the J2000 source ids have a proper specification? If yes, I can probably improve the parsing a lot. I haven't found anything on the web.
> 
> I've changed to parsing to meet the following format-specification: (J|+|-)(DD|HH)MMSS(.s+)* Is this fine?

As far as I have seen, this is not a standard format for RA and Dec. The main formats I am familiar with are described here, so I believe we need the conversion methods you wrote above. https://docs.astropy.org/en/stable/coordinates/index.html#coordinate-access

karabo/simulation/sky_model.py
Lukas113: Yeah I'm also not sure. However, I decided to put it into `SkySourcesUnits` because it's more of a helper-function not directly related to `SkyModel` and it converts the str to RA,DEC, which has something to do with units.

karabo/test/test_skymodel.py
Lukas113: I've updated the test

Lukas113:
> > @lmachadopolettivalle do you know if the J2000 source ids have a proper specification? If yes, I can probably improve the parsing a lot. I haven't found anything on the web.
> > I've changed to parsing to meet the following format-specification: (J|+|-)(DD|HH)MMSS(.s+)* Is this fine?
> 
> As far as I have seen, this is not a standard format for RA and Dec. The main formats I am familiar with are described here, so I believe we need the conversion methods you wrote above. https://docs.astropy.org/en/stable/coordinates/index.html#coordinate-access

Ok. Then I hope the conversion function is good enough for the usual use-cases. As far as I've seen this function should work at least for the gleam, mightee and mals survey, tested on example conversions for each survey.

lmachadopolettivalle (approved):

lmachadopolettivalle (approved):

Issue #457, opened 2023-07-07
Karabo tests are enhanced


kenfus:
- Every test has at least one assert to avoid that tests only check if they run or not
- There is a folder with generated images for visual inspection

PR #567, opened 2024-04-23
Conclude InterferometerSimulation consistency between OSKAR and RASCIL

lmachadopolettivalle:
- Include `RASCIL` implementation into `InterferometerSimulation`, to compute visibilities from a `SkyModel`
- Replace implementation of line emission simulation with updated code, now using unified interfaces
- Remove notebooks in favor of example within `line_emission.py`
- Make `Imager` implementations and interfaces consistent between OSKAR and RASCIL
- Use `ImageMosaicker` instead of `MontagePY` for mosaic generation
- Improve conversion of `SkyModel` into RASCIL-compatible format


karabo/test/test_line_emission.py
mpluess, line 1: If you can, add some asserts to this test (see also https://github.com/i4Ds/Karabo-Pipeline/issues/457)

mpluess (approved):
LGTM

karabo/imaging/imager.py
```diff
@@ -212,9 +218,97 @@ def __init__(
...
+        imager.run(return_images=1)
+
+        # OSKAR adds _I.fits to the fits_path set by the user
+        image = Image(path=f"{fits_path}_I.fits")
+
+        # OSKAR Imager always produces one image by
+        # combining all frequency channels.
+        # To maintain the same data shape when compared to other imagers (e.g. RASCIL),
+        # We add an axis for frequency, and modify the header accordingly
+        assert len(image.data.shape) == 4
```
Lukas113, line 265: Same comment as the first one (ValueError & ndim)

karabo/simulation/interferometer.py
```diff
@@ -367,6 +421,85 @@ def set_ionosphere(self, file_path: str) -> None:
...
+        )
+
+        # Compute visibilities from SkyComponent list using DFT
+        vis = dft_skycomponent_visibility(
+            vis, skycomponents, dft_compute_kernel="cpu_looped"
+        )
+        # Save visibilities to disk
+        export_visibility_to_hdf5(vis, self.vis_path)
+
+        return vis
```
Lukas113, line 501: Seems very nice!

karabo/simulation/line_emission.py
```diff
@@ -1,134 +1,192 @@
...
+                imaging_backend=imaging_backend,
+                combine_across_frequencies=True,
+            )
+
+            dirty_images[-1].append(dirty)
 
-    r_array, phi_array = polar_corrdinates_grid((x_len, y_len), (x_center, y_center))
+            # dirty.data.shape meaning:
+            # (frequency channels, polarisations, pixels_x, pixels_y)
+            assert len(dirty.data.shape) == 4
```
Lukas113, line 157: ndim

karabo/simulation/line_emission.py
```diff
@@ -1,134 +1,192 @@
...
+    ), f"{len(dirty_images[0])}, {len(pointings)}"
 
-    return image
+    return visibilities, dirty_images
+
+
+def polar_corrdinates_grid(
+    im_shape: Tuple[int, int], center: Tuple[float, float]
+) -> Tuple[NDArray[np.float_], NDArray[np.float_]]:
+    raise DeprecationWarning("This function is no longer available.")
```
Lukas113, line 173: Are you sure you want to raise a warning rather than an exception?

karabo/simulation/line_emission.py
```diff
@@ -1,134 +1,192 @@
...
+
+
+def circle_image(image: NDArray[np.float_]) -> NDArray[np.float_]:
+    raise DeprecationWarning("Use the circle method within the Image class instead.")
```
Lukas113, line 177: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -1,134 +1,192 @@
...
-    header["CRVAL1"] = ra_deg
-    header["CRVAL2"] = dec_deg
-    header["CDELT1"] = -cut / float(img_size)  # 1 arcsecond per pixel
-    header["CDELT2"] = cut / float(img_size)
-    header["CRPIX1"] = img_size / 2.0
-    header["CRPIX2"] = img_size / 2.0
-    header["EQUINOX"] = 2000.0
-
-    return header
+    raise DeprecationWarning("To generate image mosaics, use the ImageMosaicker class.")
```
Lukas113, line 183: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -1,134 +1,192 @@
...
-        imaging_npixel=img_size,
-        imaging_cellsize=cut / img_size,
-        imaging_dopsf=False,
-    )
-    dirty = imager.get_dirty_image()
-    dirty.write_to_file(f"{outfile}.fits", overwrite=True)
-    dirty_image = cast(NDArray[np.float_], dirty.data[0][0])
-
-    return dirty_image
+    raise DeprecationWarning("Use the Imager class, with the RASCIL imaging backend.")
```
Lukas113, line 189: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -138,33 +196,7 @@ def oskar_imager(
...
-        weighting="Uniform",
-        uv_filter_max=3000,
-    )
-    imager.set_vis_phase_centre(ra_deg, dec_deg)
-    imager.output_root = outfile
-
-    output = imager.run(return_images=1)
-    image = cast(NDArray[np.float_], output["images"][0])
-    return image
+    raise DeprecationWarning("Use the Imager class, with the OSKAR imaging backend.")
```
Lukas113, line 199: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -176,122 +208,11 @@ def plot_scatter_recon(
...
-
-    ax2 = fig.add_subplot(122, projection=wcs, slices=slices)
-    recon_img = ax2.imshow(
-        recon_image, cmap="YlGnBu", origin="lower", vmin=vmin, vmax=vmax
-    )
-    plt.colorbar(recon_img, ax=ax2, label="Flux Density [Jy]")
-
-    plt.tight_layout()
-    plt.savefig(outfile)
+    raise DeprecationWarning("Use Image.plot_side_by_side_with_skymodel() instead.")
```
Lukas113, line 211: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -176,122 +208,11 @@ def plot_scatter_recon(
...
-
-        freq_channels_array = convert_z_to_frequency(redshift_channels_array)
-
-    freq_bins = np.abs(np.diff(freq_channels_array))
-
-    print("The frequency channel starts at:", freq_start, "Hz")
-    print("The bin sizes of the freq channel are:", freq_bins, "Hz")
-
-    return redshift_channels_array, freq_channels_array, freq_bins, freq_mid
+    raise DeprecationWarning("Use sky.filter_by_column(13, z_min, z_max) instead.")
```
Lukas113, line 215: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -316,114 +237,7 @@ def karabo_reconstruction(
...
-        if verbose:
-            print(
-                "Write the reconstructed image to a fits file which can be used for "
-                "coaddition.",
-                mosaic_pntg_file,
-            )
-        fits.writeto(f"{mosaic_pntg_file}.fits", dirty_image, header, overwrite=True)
-
-    return dirty_image, header
+    raise DeprecationWarning("Use line_emission_pipeline() instead.")
```
Lukas113, line 240: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -445,64 +259,7 @@ def run_one_channel_simulation(
...
-        cut=cut,
-        img_size=img_size,
-        channel_num=1,
-        circle=circle,
-        rascil=rascil,
-        verbose=verbose,
-    )
-
-    return dirty_image, header
+    raise DeprecationWarning("Use line_emission_pipeline() instead.")
```
Lukas113, line 262: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -511,32 +268,13 @@ def create_line_emission_h5_file(
     redshift_channel: NDArray[np.float_],
     header: fits.header.Header,
 ) -> None:
+    raise DeprecationWarning(
```
Lukas113, line 271: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -558,180 +296,11 @@ def line_emission_pointing(
...
-    print("Save 3-dim reconstructed dirty images as h5")
-    create_line_emission_h5_file(
-        filename=outpath / ("line_emission_dirty_images.h5"),
-        dirty_images=dirty_images,
-        redshift_channel=redshift_channel,
-        header=header,
-    )
-
-    return dirty_image, dirty_images, header, freq_mid
+    raise DeprecationWarning("Use line_emission_pipeline() instead.")
```
Lukas113, line 299: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -558,180 +296,11 @@ def line_emission_pointing(
...
-
-    :param freq: Frequency of interest in Hz.
-    :return: The power pattern FWHM of the MeerKAT telescope at this frequency in
-             degrees.
-    """
-    root = cast(np.float64, np.sqrt(89.5 * 86.2))
-    gaussian_fwhm = root / 60.0 * (1e3 / (freq / 10**6))
-
-    return gaussian_fwhm
+    raise DeprecationWarning("This function has been removed.")
```
Lukas113, line 303: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -742,37 +311,7 @@ def gaussian_beam(
...
-    beam_image = gauss_kernel.array
-    # normalize the kernel, such that max=1.0
-    beam_image = beam_image / np.max(beam_image)
-
-    # make the beam image circular and save it as a fits file
-    beam_image = circle_image(beam_image)
-    fits.writeto(outfile, beam_image, header, overwrite=True)
-
-    return beam_image, header
+    raise DeprecationWarning("This function has been removed.")
```
Lukas113, line 314: Same comment as above with warning vs exception.

karabo/simulation/line_emission.py
```diff
@@ -784,39 +323,187 @@ def simple_gaussian_beam_correction(
...
-    outpath = Path(outpath)
-    print("Calculate gaussian beam for primary beam correction...")
-    beam, header = gaussian_beam(
-        img_size=img_size,
-        ra_deg=ra_deg,
-        dec_deg=dec_deg,
-        cut=cut,
-        fwhm=gaussian_fwhm,
-        outfile=outpath / "gaussian_beam.fits",
+    raise DeprecationWarning("This function has been removed.")
```
Lukas113, line 326: Same comment as above with warning vs exception.

karabo/simulation/telescope.py
```diff
@@ -142,10 +153,54 @@ def __init__(
...
+        backend: SimulatorBackend = SimulatorBackend.OSKAR,
+    ) -> Telescope:
+        ...
+
+    @classmethod
+    def constructor(
+        cls,
+        name: Union[
+            str, OSKARTelescopesWithVersionType, OSKARTelescopesWithoutVersionType
+        ],
```
Lukas113, line 203: I don't see a reason why `str` should ever be allowed, since it raises always a `ValueError`. And for the case of `name` and `version`, what is when allowed, I think it would be best so solve it with overloads. Then `asserts` would be fine in my opinion because it would be a violation of the signature. Otherwise I'd prefer to raise RuntimeError.

Lukas113 (changes requested):
Changes are looking very nice!

Just a few comments to address. Don't see them as requests, but as suggestions. I'll approve the PR once you've gone through the comments.

karabo/simulation/telescope.py
lmachadopolettivalle: These are very good points. I have introduced a list of accepted `RASCIL` telescopes to replace the `str` in the signature. And I believe the overloads already make it clear which combinations of parameters are allowed, but please let me know if that's not the case.

karabo/simulation/line_emission.py
Lukas113: Yeah I know. I've realized that just at a later point. The way I know `DeprecationWarnings` are used is to print it with `warnings.warn` a few releases earlier. However, I haven't seen any raise Warning ever elsewhere, therefore I'm not sure if this is the way to go.

I just think we should do the same as the python-community handles such things. And as far as I've seen, there is no such error-handling after the removal of api-functions. Just warn with `DeprecationWarnings` a few releases earlier.

Therefore, my suggestion is to just remove the functions. However, this is just a suggestion. If you feel otherwise and have a reason to do so, go for it.

mpluess (approved):
LGTM

karabo/imaging/imager.py
```diff
@@ -223,39 +317,75 @@ def get_dirty_image(
...
-        return image
+            return image
+        elif imaging_backend is SimulatorBackend.RASCIL:
+            vis = self.visibility
+            if isinstance(vis, Visibility):
+                # Convert OSKAR Visibility to RASCIL-compatible format
+                block_visibilities = create_visibility_from_ms(str(vis.ms_file_path))
+
+                if len(block_visibilities) != 1:
+                    raise EnvironmentError("Visibilities are too large")
```
Lukas113, line 378: to me it seems like this is more a `ValueError` rather than an `EnvironmentError` (see [EnvironmentError](https://www.geeksforgeeks.org/environmenterror-exception-in-python/))

karabo/simulation/telescope.py
```diff
@@ -142,10 +171,58 @@ def __init__(
...
-        name: Union[OSKARTelescopesWithVersionType, OSKARTelescopesWithoutVersionType],
+        name: Union[
+            RASCILTelescopes,
+            OSKARTelescopesWithVersionType,
+            OSKARTelescopesWithoutVersionType,
+        ],
+        version: Optional[enum.Enum] = None,
+        backend: SimulatorBackend = SimulatorBackend.OSKAR,
+    ) -> Telescope:
+        ...
```
Lukas113, line 216: To me it seems like this is not a valid signature overload (according to the code), but all above are.

karabo/simulation/telescope.py
```diff
@@ -166,35 +243,38 @@ def constructor(
...
+                ), f"""version is not a required field
+                    for telescope {name}, but {version} was provided.
                     Please do not provide a value for the version field."""
-                    )
+            else:
+                raise ValueError(
+                    f"""
+                    {name = } is not an accepted telescope name for this backend.
+                """
+                )
```
Lukas113, line 277: In this case, this is either a `TypeError` or an `assert_never` if you would perform the type-narrowing using a `TypeGuard` function. However, it's way easier to just raise a `TypeError` and not have to care about `TypeGuard`s.

Lukas113: In addition, I've just seen that on line 309-311 is an `assert_never` and not a `ValueError` (since it's easy to just change that)

Lukas113 (changes requested):
LGTM, just some minor requests.

karabo/simulation/telescope.py
```diff
@@ -166,35 +243,38 @@ def constructor(
...
-                if (version is None) or (version not in accepted_versions):
-                    raise ValueError(
-                        f"""{version} is not valid for telescope {name}.
-                        List of valid versions: {accepted_versions}."""
-                    )
+                assert (
+                    version is not None
+                ), f"""version is a required field
+                    for telescope {name}, but was not provided.
+                    Please provide a value for the version field."""
```
Lukas113, line 258: Just a quick reminder. using triple quotes and newline also results in a newline on the output. Because the sentence is not finished after e.g. "field", it's probably not intended? As mentioned, this is just a reminder, and because this not a huge thing and I don't request it to change anywhere.

PR #569, opened 2024-05-17
WSClean integration, new imaging interface

### LINKED ISSUE #123: Implement WSCLEAN for the deconvolution
The RASCIL imager is providing a slow and inconsistent. It would be important for both the accuracy and easy usage to integrate WSCLEAN in Karabo.
https://gitlab.com/aroffringa/wsclean/-/tree/master/

mpluess:
Closes #123


karabo/imaging/imager_oskar.py
```diff
@@ -0,0 +1,149 @@
...
+
+        If config is a DirtyImagerConfig (base class) instance, converts it to
+        an OskarDirtyImagerConfig using the
+        OskarDirtyImagerConfig.from_dirty_imager_config method.
+
+        Args:
+            config (DirtyImagerConfig): see config attribute
+        """
+        if not isinstance(config, OskarDirtyImagerConfig):
+            config = OskarDirtyImagerConfig.from_dirty_imager_config(config)
```
Lukas113, line 78: I don't like this design choice. It assumes that all sub-configs of `DirtyImagerConfig` now and in the future have default arguments.

karabo/imaging/imager_oskar.py
```diff
@@ -0,0 +1,149 @@
...
+        self,
+        visibility: Union[Visibility, RASCILVisibility],
+        output_fits_path: Optional[FilePathType] = None,
+    ) -> Image:
+        if isinstance(visibility, RASCILVisibility):
+            raise NotImplementedError(
+                """OSKAR Imager applied to
+                RASCIL Visibilities is currently not supported.
+                For RASCIL Visibilities please use the RASCIL Imager."""
+            )
```
Lukas113, line 92: You can also support this by removing the `RASCILVisibility` union type from the signature.

karabo/imaging/imager_oskar.py
```diff
@@ -0,0 +1,149 @@
...
+
+        config: OskarDirtyImagerConfig = cast(OskarDirtyImagerConfig, self.config)
```
Lukas113, line 94: This shouldn't be necessary. It's because you don't set self.config in the sub-class, but in the abstract super-class.

karabo/imaging/imager_oskar.py
```diff
@@ -0,0 +1,149 @@
...
+            phase_centre = SkyCoord(config.imaging_phasecentre, frame="icrs")
+            ra = phase_centre.ra.degree
+            dec = phase_centre.dec.degree
+
+            imager.set_vis_phase_centre(ra, dec)
+
+        imager.run(return_images=1)
+
+        # OSKAR adds _I.fits to the fits_path set by the user
+        image = Image(path=f"{output_fits_path}_I.fits")
```
Lukas113, line 136: I think there is a mistake here. Most tools I know assume that you add the ending (.fits in this case) yourself. Even if not, above on line 102, the ending is also added if not provided. This results in a weird file-ending.

karabo/imaging/imager_oskar.py
```diff
@@ -0,0 +1,149 @@
...
+        # OSKAR Imager always produces one image by
+        # combining all frequency channels.
+        # To maintain the same data shape when compared to other imagers (e.g. RASCIL),
+        # We add an axis for frequency, and modify the header accordingly
+        assert image.data.ndim == 4
+
+        image.header["NAXIS"] = 4
+        image.header["NAXIS4"] = 1
+
+        image.write_to_file(path=f"{output_fits_path}_I.fits", overwrite=True)
```
Lukas113, line 147: Same comment as above.

Lukas113 (changes requested):

karabo/imaging/imager_base.py
```diff
@@ -0,0 +1,152 @@
...
+    """Base class for the config / parameters of a dirty imager.
+
+    Contains basic parameters common across all dirty imagers.
+    Inherit and add parameters specific to a dirty imager implementation.
+
+    Attributes:
+        imaging_npixel (int): Image size
+        imaging_cellsize (float): Scale of a pixel in radians
+        combine_across_frequencies (bool): Whether or not to combine images
+            across all frequency channels into one image. Defaults to True.
```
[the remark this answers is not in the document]
mpluess: Makes sense. That's just how VS Code generated the docstring skeleton, so I didn't bother deleting the types.

karabo/imaging/imager_base.py
```diff
@@ -0,0 +1,152 @@
...
+    ) -> Image:
+        """Creates a clean image from a dirty image or from visibilities.
+
+        Args:
+            ms_file_path (Optional[FilePathType]): Path to measurement set from which
+                a clean image should be created. Provide either ms_file_path or
+                dirty_fits_path, not both, not none. Defaults to None.
+            dirty_fits_path (Optional[FilePathType]): Path to dirty image FITS file from
+                which to create a clean image. Provide either ms_file_path or
+                dirty_fits_path, not both, not none. Defaults to None.
```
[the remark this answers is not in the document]
mpluess: Good point. Since this interface is not final and will probably be changed again in the near future (more discussions amongst ourselves and ideally user input needed), I would suggest leaving it as-is for now.

karabo/imaging/imager_oskar.py
```diff
@@ -0,0 +1,149 @@
...
+
+    Attributes:
+        imaging_npixel (int): see DirtyImagerConfig
+        imaging_cellsize (float): see DirtyImagerConfig
+        combine_across_frequencies (bool): see DirtyImagerConfig
+        imaging_phasecentre (Optional[str]): Phase centre (in SkyCoord string format).
+            Defaults to None.
+    """
+
+    imaging_phasecentre: Optional[str] = None
```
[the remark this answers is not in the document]
mpluess: Wow, that's a good point, I didn't think about that.
By coincidence, the way I designed the whole thing, you should provide defaults for all attributes in extended configs anyway. That makes it possible to pass a base config, e.g. DirtyImagerConfig, which is then converted to a config for the specific imager, e.g. OskarDirtyImagerConfig, using the values from the passed base config plus defaults for OSKAR-specific attributes.
I would therefore leave it as is.

karabo/imaging/imager_oskar.py
mpluess: Closely related to my comment above. Let's discuss this tomorrow.

karabo/imaging/util.py
```diff
@@ -0,0 +1,293 @@
...
+    -------
+    List[SkaSdpImage]
+        List of images from the MGCLS Enhanced Products bucket.
+    """
+    mgcls_cdo = MGCLSContainerDownloadObject(regexr_pattern=regex_pattern)
+    local_file_paths = mgcls_cdo.get_all(verbose=verbose)
+    if len(local_file_paths) == 0:
+        raise ValueError(
+            f"No files in {mgcls_cdo._remote_container_url} for {regex_pattern=}"
+        )
```
Lukas113, line 133: I think `FileNotFoundError` is a better choice here. Even though I think you haven't written this function by yourself, just moved it into this file.

karabo/test/test_imager_wsclean.py
```diff
@@ -0,0 +1,212 @@
...
+def test_create_cleaned_image_custom_path():
+    visibility = _run_sim()
+
+    imaging_npixel = 2048
+    imaging_cellsize = 3.878509448876288e-05
+
+    output_fits_path = os.path.join(
+        FileHandler().get_tmp_dir(prefix="test_imager_wsclean-"),
+        "test_create_cleaned_image_custom_path.fits",
+    )
```
Lukas113, line 130: I think it's a better choice to use tmpfile or dir instead of the FileHandler to not leave directories around after testing. The FileHandler is supposed to be used for disk-caching which should remain a certain amount of time on disk. Otherwise, you can also just use `with FileHandler()`, which should also work.

Lukas113 (changes requested):

karabo/imaging/imager_oskar.py
mpluess: This is code I didn't touch and I think Luis was recently battling with the OSKAR FITS paths to make it work.
@lmachadopolettivalle Do you think there's room for improvement here? See also comment below.

karabo/imaging/imager_base.py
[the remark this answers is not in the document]
Lukas113: Sure :thumbsup:, as long as it doesn't get merged like this

karabo/imaging/imager_base.py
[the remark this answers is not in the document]
Lukas113: Yeah it makes the same for me :smile:. I haven't seen any option to not do this in VS Code.

karabo/imaging/imager_rascil.py
```diff
@@ -0,0 +1,546 @@
...
+                purpose="disk-cache for dirty.fits",
+            )
+            output_fits_path = os.path.join(tmp_dir, "dirty.fits")
+
+        if isinstance(visibility, Visibility):
+            # Convert OSKAR Visibility to RASCIL-compatible format
+            block_visibilities = create_visibility_from_ms(str(visibility.ms_file_path))
+
+            if len(block_visibilities) != 1:
+                raise ValueError("Visibilities are too large")
```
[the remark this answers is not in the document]
mpluess: @lmachadopolettivalle I think this is your field of expertise ;)

karabo/imaging/imager_oskar.py
lmachadopolettivalle: For this comment and the comment below, this is a quirk of how OSKAR stores the .fits files it generates. OSKAR introduces the ending `_I.fits`, and we account for this when we read the data here, and when we overwrite the data below.

karabo/imaging/imager_oskar.py
lmachadopolettivalle: See comment above about the fits file ending from OSKAR.

karabo/imaging/imager_rascil.py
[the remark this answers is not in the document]
lmachadopolettivalle: You are right, the most general usage of `create_visibility_from_ms` could return a list with more than one visibility. The check for length == 1 is based on how we have set up our pipelines (e.g. every `Visibility` object only contains one set of visibilities. One idea is to remove this `if` block, continue using the 0th entry of the list, and warn the user in case the length is greater than 1.

It'd take a bit more care for us to use every entry of the list and combine the visibilities in some meaningful way, so I think the approach above is a good first step towards making this code more general.

karabo/imaging/imager_oskar.py
mpluess: seems a minor / cosmetic thing to me as long as it works. let's either open a low-prio issue or leave it for the time being.

karabo/imaging/imager_wsclean.py
```diff
@@ -0,0 +1,320 @@
...
+
+        tmp_dir = FileHandler().get_tmp_dir(
+            prefix=self.TMP_PREFIX_CLEANED,
+            purpose=self.TMP_PURPOSE_CLEANED,
+        )
+        prefix = str(uuid.uuid4())
+        if dirty_fits_path is not None:
+            shutil.copyfile(
+                dirty_fits_path,
+                os.path.join(tmp_dir, f"{prefix}-dirty.fits"),
```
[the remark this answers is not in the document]
mpluess: It is necessary to set a prefix when continuing from a dirty image due to how wsclean works (CLI parameter -reuse-dirty <prefix>). You're right that it's not necessary to use a UUID as a prefix here. I changed it to a fixed prefix "pre_existing".

karabo/imaging/imager_rascil.py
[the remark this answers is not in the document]
mpluess: is a warning strong enough? maybe a NotImplementedError("This imager currently doesn't support more than one visibility.")?

karabo/imaging/imager_oskar.py
Lukas113: Hmm, but then the line 77 is incorrect, because there, a `dirty.fits` is set as the file-name, which results in a double-ending file-name.

In addition, I don't really like this behavior, especially if a path is provided explicitly. What I'd like most is to rename the resulting file to it's intended name. Everything else seems inconsistent to me.

I agree it's not a huge thing. To me it seems like it's almost no effort to change it.

karabo/test/test_imager_wsclean.py
```diff
@@ -0,0 +1,190 @@
...
+
+    dirty_imager = WscleanDirtyImager(
+        DirtyImagerConfig(
+            imaging_npixel=2048,
+            imaging_cellsize=3.878509448876288e-05,
+        ),
+    )
+    output_fits_path = os.path.join(
+        FileHandler().get_tmp_dir(prefix="test_imager_wsclean-"),
+        "test_dirty_image_custom_path.fits",
```
Lukas113, line 46: I think replacing FileHandler wich tmpdir or use it with `with` to not leaving a test-cache directory behind is a better choice.

karabo/imaging/imager_oskar.py
mpluess: You're right and that's a good solution. I implemented it this way and added a test to make sure passing a custom output path works as expected.

Lukas113 (approved):
LGTM :thumbsup:

PR #590, opened 2024-07-14
585 obscore

### LINKED ISSUE #585: IVOA ObsCore Metadata
Karabo should support creating ObsCore metadata from the simulator to enable data-product integration into ObsTAP services. This is a relevant use-case for the SRCNet efforts.

ObsCore documentation [here](https://ivoa.net/documents/ObsCore/)
ObsCore key-values overview: p.15/16, full-overview (incl. non-mandatory) p.40-43.

Visibilities and image files therefore need an according metadata (JSON) file.

Lukas113:
closes #585 

Supports [ObsCore-v1.1](https://ivoa.net/documents/ObsCore/) metadata creation and some checks for simulated data of Karabo.

This is a feature towards the SRCNet efforts to create simulated data & attach the needed metadata to it. I've only considered fields which are currently supported by the current SRCNet Rucio service.


Lukas113:
Tests should succeed as soon as #591 is merged into main and this branch is updated.

karabo/data/obscore.py
```diff
@@ -0,0 +1,876 @@
...
+        target_name: Astronomical object observed, if any. This is typically the name
+            of an astronomical object, but could be the name of a survey field.
+
+        s_ra: [deg] Central right ascension, ICRS.
+
+        s_dec: [deg] Central declination, ICRS.
+
+        s_fov: [deg] Region covered by the data product. For a circular region, this
+            is the diameter. For most data products, the value should be large enough
+            to include the entire are of the observation. For detailed spatial
```
mpluess, line 219: Typo: are -> area

karabo/data/obscore.py
```diff
@@ -0,0 +1,876 @@
...
+            ocm.t_resolution = t_res
+            num_elements_t = int(total_duration_sec / t_res)
+            ocm.t_xel = num_elements_t
+            freq_start_hz = header.freq_start_hz  # midpoint freq of first channel
+            channel_bandwidth_hz = header.channel_bandwidth_hz
+            freq_inc_hz = header.freq_inc_hz
+            min_freq_hz = freq_start_hz - channel_bandwidth_hz / 2
+            n_channels = header.num_channels_total
+            max_freq_hz = min_freq_hz + freq_inc_hz * n_channels
+            c = 299792458  # m/s
```
mpluess, line 467: Getting this value from astropy seems cleaner to me. https://docs.astropy.org/en/stable/constants/index.html

karabo/data/obscore.py
```diff
@@ -0,0 +1,876 @@
...
+    def from_visibility(
+        cls,
+        vis: Visibility,
+        calibrated: Optional[bool] = None,
+        tel: Optional[Telescope] = None,
+        obs: Optional[Observation] = None,
+    ) -> Self:
+        """Suggests fields from `Visibility`.
+
+        This function may not adjust each field for your needs. In addition, there
```
mpluess, line 422: Mentioning which mandatory fields will not be set would be nice.

karabo/data/obscore.py
```diff
@@ -0,0 +1,876 @@
...
+        n_channels = fits_axes.freq.naxis(header=header)
+
+        if (s_pixel_scale := abs(x_inc_deg)) != abs(y_inc_deg):
+            wmsg = (
+                f"Pixel-size is not square for `s_pixel_scale`: {x_inc_deg=}, "
+                + f"{y_inc_deg}. `s_pixel_scale` set to {s_pixel_scale=}."
+            )
+            warn(message=wmsg, category=UserWarning, stacklevel=1)
+        min_freq_hz = freq_center_hz - freq_inc_hz / 2
+        c = 299792458  # m/s
```
mpluess, line 547: Getting this value from astropy seems cleaner to me. https://docs.astropy.org/en/stable/constants/index.html

karabo/data/obscore.py
```diff
@@ -0,0 +1,876 @@
...
+        elif pol_xel is not None and pol_states is None:
+            valid = False
+            if verbose:
+                wmsg = (
+                    f"{pol_xel=} is specified, but {pol_states=} which isn't consistent"
+                )
+                warn(message=wmsg, category=UserWarning, stacklevel=1)
+        elif pol_xel is not None and pol_states is not None:
+            if pol_xel != (num_pol_states := len(pol_states)):
+                wmsg = f"{pol_xel=} should be {num_pol_states=}"
```
mpluess, line 841: valid is not set to False here, neither is a warning issued. Is this on purpose?

karabo/simulation/telescope.py
```diff
@@ -748,3 +765,67 @@ def create_baseline_cut_telelescope(
...
+        B: Max baseline in meters
+
+        Args:
+            freq: Frequency [Hz].
+            b: Max baseline in meters (e.g. from `longest_baseline`).
+
+        Returns:
+            Angular resolution in arcsec.
+        """
+        ang_res = (299792458 / freq) / b * u.deg
```
mpluess, line 829: Getting this value from astropy seems cleaner to me. https://docs.astropy.org/en/stable/constants/index.html

karabo/data/obscore.py
```diff
@@ -0,0 +1,876 @@
...
+    @classmethod
+    def from_image(
+        cls,
+        img: Image,
+        fits_axes: FitsHeaderAxes = FitsHeaderAxes(),  # immutable default
+    ) -> Self:
+        """Update fields from `Image`.
+
+        This function may not adjust each field for your needs. In addition, there
+            is no possibility to fill some mandatory fields because there is just no
```
mpluess, line 506: Mentioning which mandatory fields will not be set would be nice.

mpluess (changes requested):
Nice work overall!

karabo/data/obscore.py
```diff
@@ -0,0 +1,876 @@
...
+            header: Header to extract unit from.
+
+        Returns:
+            Unit as str.
+        """
+        return str(header[f"CTYPE{self.axis}"])
+
+
+@dataclass(frozen=True)
+class FitsHeaderAxes:
```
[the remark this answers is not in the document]
Lukas113: Good catch! that was a bug I only fixed in code XD

karabo/data/obscore.py
Lukas113: Hmm, not sure about that. Mandatory fields are already depicted in the class docstring and when creating a JSON. One of the mandatory field is also just set depending if an input is set or not.

karabo/data/obscore.py
Lukas113: Hmm, maybe. c is a constant integer since the redefinition of the meter, thus should not be changed. However, if it would change again :), then this would be auto corrected. On the other hand, you introduce another module dependency which could lead to an API breaking change. Especially since astropy.constants.c.value is not properly typed.

Long story short, I think both arguments are valid. The probably cleanest way to do it is to define our one constants, where they may or may not come from astropy and import it from there. However, I'm too lazy for that XD. I just take the astropy constant.

.gitignore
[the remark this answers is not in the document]
Lukas113: Not sure from where. The .tm files should be read-only. Maybe a bug?

karabo/data/obscore.py
Lukas113: Not sure about that for the same reason mentioned in `from_visibility`

.gitignore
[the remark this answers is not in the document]
mpluess: I have no idea what the purpose of these .bin files is. I just thought that this may be related to you removing the `karabo/data/**` line from .gitignore. Maybe the .bin files were the reasons for this line in the first place.

karabo/data/obscore.py
mpluess: Alright, let's leave it as is.

.gitignore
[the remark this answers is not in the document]
Lukas113: To me this doesn't make any sense. Having it in the .gitignore means that there may be some created files. And according to my understanding this shouldn't happen. Anyway, out of scope for this PR I guess :)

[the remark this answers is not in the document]
mpluess: Okay, let's leave it for now

mpluess (approved):
LGTM

PR #591, opened 2024-07-16
addressed mypy issues due to matplotlib update :clock430:

Lukas113:
mypy updated their type-hints which lead to mypy breaking changes (not API).

Our CI-tests were failing due to this issue and this stopped other PR to be accepted.


mpluess (approved):
LGTM

Issue #583, opened 2024-06-27
Improve Doc


Lukas113:
- [ ] Doc how to do a PR
- [ ] Chose docstringstyle which is testable through e.g. `ruff`
- [ ] Improve docstrings
- [ ] Improve documentation overall (code should/must be testable)

PR #592, opened 2024-07-16
SRCNet demo notebook, updated examples page

mpluess:
JIRA ticket CHOC-9


.pre-commit-config.yaml
```diff
@@ -4,6 +4,7 @@ repos:
     hooks:
     -   id: check-added-large-files
         additional_dependencies: [tomli]
+        exclude: ".*.ipynb$"
```
Lukas113, line 7: I see that you don't want have jupyter notebooks in here. However, I think it's a better idea to limit it to just the notebooks in the `karabo/examples/` directory.

In addition, I found a way to [exclude jupyter-notebooks](https://github.com/OpenEnergyPlatform/open-MaStR/pull/296/files) from the language section. I think just changeing the `postprocessing/**` to `karabo/examples/**` should do the trick. They just distort the repository language structure. If you agree, I think you can just put this into this PR.

doc/src/examples/example_structure.md
Lukas113, line 1: Hmm, isn't the `example_structure.md` just a partial copy of `examples.md`? If yes, I wonder what the reason for this is?  I think we can just get rid of `example_structure.md` because of duplication? Do you agree?

Lukas113: I'm also wondering if the code in the examples is up to date. Do you know or can you test it? If it's not, I think we should open a new issue if it not just a 5min fix to change it.

Lukas113: Ideally we should just have testable code examples, which is doable according to my knowledge. However, I see that's not really in the scope of your PR.

karabo/examples/SRCNet_simulation_walkthrough.ipynb
Lukas113, line 1: Notebook looks great :+1: 

I think adding some meaningful sections would even improve the notebook further.

Lukas113 (changes requested):

doc/src/examples/example_structure.md
Lukas113: Ok, thanks. Didn't know that. I think the testable code examples should already be covered by issue #583 . I just wanted to be sure that what's now in the doc is valid.

karabo/examples/SRCNet_simulation_walkthrough.ipynb
mpluess: Good point, added sections

Lukas113 (approved):
LGTM

PR #594, opened 2024-07-17
Script to generate SRCNet v0.1 test data

mpluess:
JIRA ticket CHOC-7


karabo/examples/SRCNet_v0.1_simulation.py
Lukas113, line 1: I think the following makes sense:

First, put launch-code into:
```python
if __name__ == "__main__":
    main()
```

Secondly, add argparse args for configurable parameters? Not sure about this one because it's just an example script. Depends on how you thing this should be used I think.

Lukas113 (changes requested):

karabo/examples/SRCNet_v0.1_simulation.py
mpluess: 1. Good catch
2. The idea is for the script to be run as-is, not changing any parameters. Therefore, argparse seems unnecessary to me.

Lukas113 (approved):
LGTM

PR #598, opened 2024-07-24
Download timeout, improved docs

mpluess:
Fixes issues that popped up during https://jira.skatelescope.org/browse/CHOC-21


Lukas113 (approved):
LGTM

Maybe a shorter timeout is even better? Not sure tough.

mpluess:
> LGTM
> 
> Maybe a shorter timeout is even better? Not sure tough.

possibly, thought i'd rather be on the safe side for now.

Issue #580, opened 2024-06-24
Exception raised when saving telescope


anawas:
# Summary
When you try to write a telescope configuration with `telescope.write_to_file("mytelescope.tm")` you get an exception saying, that the file `position.txt` inside folder `mytelescope.tm`does not exist. You can create an empty file and the code runs with no errors.

# How to reproduce
Before running the code make sure there is no folder `mytelescope.tm`.
```
from karabo.simulation.telescope import Telescope
from karabo.simulator_backend import SimulatorBackend

SIMULATOR_BACKEND = SimulatorBackend.OSKAR
telescope = Telescope.constructor("MeerKAT", backend=SIMULATOR_BACKEND)
telescope.write_to_file("mytelescope.tm")
```
# Result
```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
Cell In[3], [line 1](vscode-notebook-cell:?execution_count=3&line=1)
----> [1](vscode-notebook-cell:?execution_count=3&line=1) telescope.write_to_file("mytelescope.tm")

File ~/miniconda3/envs/karabo/lib/python3.9/site-packages/karabo/simulation/telescope.py:472, in Telescope.write_to_file(self, dir)
    [467](https://vscode-remote+wsl-002bdebian.vscode-resource.vscode-cdn.net/home/andreas/karabo/telescope_example/~/miniconda3/envs/karabo/lib/python3.9/site-packages/karabo/simulation/telescope.py:467) def write_to_file(self, dir: DirPathType) -> None:
    [468](https://vscode-remote+wsl-002bdebian.vscode-resource.vscode-cdn.net/home/andreas/karabo/telescope_example/~/miniconda3/envs/karabo/lib/python3.9/site-packages/karabo/simulation/telescope.py:468)     """
    [469](https://vscode-remote+wsl-002bdebian.vscode-resource.vscode-cdn.net/home/andreas/karabo/telescope_example/~/miniconda3/envs/karabo/lib/python3.9/site-packages/karabo/simulation/telescope.py:469)     Create .tm telescope configuration at the specified path
    [470](https://vscode-remote+wsl-002bdebian.vscode-resource.vscode-cdn.net/home/andreas/karabo/telescope_example/~/miniconda3/envs/karabo/lib/python3.9/site-packages/karabo/simulation/telescope.py:470)     :param dir: directory in which the configuration will be saved in.
[...]

PR #601, opened 2024-07-29
Fix issue #580

anawas:
Telescope configuration can now be saved without raising an exception. However, an exception is raised in case the folder exists.

### DIFF
--- a/karabo/simulation/telescope.py
+++ b/karabo/simulation/telescope.py
@@ -54,7 +54,7 @@ from karabo.simulation.telescope_versions import (
 from karabo.simulator_backend import SimulatorBackend
 from karabo.util._types import DirPathType, NPFloatLike
 from karabo.util.data_util import get_module_absolute_path
-from karabo.util.file_handler import FileHandler
+from karabo.util.file_handler import FileHandler, write_dir
 from karabo.util.math_util import long_lat_to_cartesian
 
 OSKARTelescopesWithVersionType = Literal[
@@ -472,10 +472,8 @@ but was not provided. Please provide a value for the version field."
             mkdir=False,
         )
         tmp_dir = os.path.join(tmp_dir, "oskar-telescope")
-        os.makedirs(tmp_dir, exist_ok=True)
-        if not FileHandler.is_dir_empty(dirname=tmp_dir):
-            FileHandler.empty_dir(dir_path=tmp_dir)
-        self.write_to_file(tmp_dir)
+        if not os.path.exists(tmp_dir):
+            self.write_to_file(tmp_dir)
         tel = OskarTelescope()
         tel.load(tmp_dir)
         self.path = tmp_dir
@@ -485,22 +483,22 @@ but was not provided. Please provide a value for the version field."
         """
         Create .tm telescope configuration at the specified path
         :param dir: directory in which the configuration will be saved in.
-        :param overwrite: If True an existing directory is overwritten.
+        :param overwrite: If True an existing directory is overwritten if exists. Be
+            careful to put the correct dir as input because the old one can get removed!
         """
-        if not FileHandler.is_dir_empty(dirname=dir):
-            FileHandler.empty_dir(dir_path=dir)
-        self.__write_position_txt(os.path.join(dir, "position.txt"))
-        self.__write_layout_txt(
-            os.path.join(dir, "layout.txt"),
-            [station.position for station in self.stations],
-        )
-        for i, station in enumerate(self.stations):
-            station_path = f"{dir}{os.path.sep}station{'{:03d}'.format(i)}"
-            os.mkdir(station_path)
+        with write_dir(dir=dir, overwrite=overwrite) as wd:
+            self.__write_position_txt(os.path.join(wd, "position.txt"))
             self.__write_layout_txt(
-                os.path.join(station_path, "layout.txt"),
-                station.antennas,
+                os.path.join(wd, "layout.txt"),
+                [station.position for station in self.stations],
             )
+            for i, station in enumerate(self.stations):
+                station_path = f"{wd}{os.path.sep}station{'{:03d}'.format(i)}"
+                os.mkdir(station_path)
+                self.__write_layout_txt(
+                    os.path.join(station_path, "layout.txt"),
+                    station.antennas,
+                )
 
     def __write_position_txt(self, position_file_path: str) -> None:
         position_file = open(position_file_path, "a")
--- a/karabo/simulation/telescope.py
+++ b/karabo/simulation/telescope.py
@@ -481,7 +481,7 @@ but was not provided. Please provide a value for the version field."
         self.path = tmp_dir
         return tel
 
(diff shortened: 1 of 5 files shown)


karabo/simulation/telescope.py
```diff
@@ -475,17 +475,21 @@ def get_OSKAR_telescope(self) -> OskarTelescope:
...
-    def write_to_file(self, dir: DirPathType) -> None:
+    def write_to_file(self, dir: DirPathType, *, overwrite: bool = False) -> None:
         """
         Create .tm telescope configuration at the specified path
         :param dir: directory in which the configuration will be saved in.
+        :param overwrite: If True an existing directory is overwritten.
         """
+        os.makedirs(dir, exist_ok=overwrite)
+        if not FileHandler.is_dir_empty(dirname=dir):
+            FileHandler.empty_dir(dir_path=dir)
```
Lukas113, line 492: Nice addition.

I'm also thinking about making writing files as directories more safe in case there is any interrupt before the creation has finished. Here's a suggestion:

```python
import os
import shutil
from collections.abc import Generator
from contextlib import contextmanager
from pathlib import Path
from tempfile import TemporaryDirectory
from typing import TypeVar

from karabo.util._types import DirPathType
from karabo.util.helpers import get_rnd_str

TDirPathType = TypeVar("TDirPathType", bound=DirPathType)

@contextmanager
def write_dir(dir: TDirPathType, *, overwrite: bool = False) -> Generator[TDirPathType, None, None]:
    """Enables transactional creating and writing into `dir`.

    Assumes that `dir` is empty, meaning NOT partially filled with anything.

    This function is NOT thread-safe!

    Args:
        dir: Directory to create & write.
        overwrite: Allow overwrite? Will remove the entire old `dir` if exists.

    Yields:
        Directory to fill safely.
    """
    dir_path = Path(dir)  # `Path(dir)` handles trailing "/"
    parent_dir, dirname = os.path.split(dir_path)
    if not os.path.exists(parent_dir):  # to not have to care about parent-dir removal if created
        err_msg = f"{parent_dir=} doesn't exist."
        raise OSError(err_msg)  # should be dev-error only
    if not overwrite and os.path.exists(dir_path):
        err_msg = f"{dir} already exists."
        raise FileExistsError(err_msg)
    rnd_str = get_rnd_str(k=10, seed=dirname)
    tmpdir = os.path.join(parent_dir, f".{rnd_str}-{dirname}")
    try:
        os.makedirs(tmpdir, exist_ok=False)
        yield tmpdir
        if overwrite and os.path.exists(dir_path):
            shutil.rmtree(path=dir_path)  # dangerous if `dir` is e.g. $HOME
        os.rename(src=tmpdir, dst=dir_path)
    except BaseException:
        shutil.rmtree(path=tmpdir)
        raise


def my_cache_fun(dir: DirPathType, *, overwrite: bool = False) -> None:
    os.makedirs(os.path.dirname(dir), exist_ok=True)
    with write_dir(dir=dir, overwrite=overwrite) as wd:
        os.mkdir(os.path.join(wd, "subdir"))
    assert os.path.exists(os.path.join(dir, "subdir"))
    print("test1 success")
    shutil.rmtree(dir)
    try:
        with write_dir(dir=dir, overwrite=overwrite) as wd:
            os.mkdir(os.path.join(wd, "subdir"))
            raise RuntimeError()
    except RuntimeError:
        pass
    assert not os.path.exists(os.path.join(dir, "subdir"))
    print("test2 success")

with TemporaryDirectory() as tmpdir:  # just to not fill disk
    my_cache_fun(dir=os.path.join(tmpdir, "test"))
```

What do you think about it?

Lukas113 (changes requested):

karabo/simulation/telescope.py
Lukas113: I for instance am not sure about he overwrite flag because it's potentially dangerous.

karabo/simulation/telescope.py
```diff
@@ -475,17 +475,21 @@ def get_OSKAR_telescope(self) -> OskarTelescope:
         os.makedirs(tmp_dir, exist_ok=True)
         if not FileHandler.is_dir_empty(dirname=tmp_dir):
             FileHandler.empty_dir(dir_path=tmp_dir)
-        self.write_to_file(tmp_dir)
+        self.write_to_file(tmp_dir, overwrite=True)
```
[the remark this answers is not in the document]
anawas: Agreed, the flag is not necessary at this line. I removed it.

karabo/simulation/telescope.py
anawas: I see your point about writing to a directory. It makes sense because in a multithreaded environment we may be writing a lot of large files into it (see measurement set). But the function has a lot of code. Where would we put it?

Regarding the flag: I would prefer a flag. The alternative is that the code throws an exception if the folder exists. This can be annoying if it happens after a long calculation. Or if I am developing a new pipeline. I have to remove the folder on every restart - or come up with a new name.

Lukas113: The function seems thread-safe if the threads are launched inside the `with`, but not if the threads are launched before. It's more about not having to care too much when to remove a directory if overwrite & some cleanup. I thinks this function can go to the `file_handler` module with some minimal tests as shown in `my_cache_fun`.

I agree, not having an `overwrite` flag can be annoying. It is more that we are aware that it can be dangerous and we shouldn't expose such a function to a user-api easily. A safe way (I can think of) is to protect a user from such a use-case is by checking the ending of the directory you know of, in your use-case a `.tm` dir, or throw an exception if `overwrite` and the ending is not a `.tm` dir. It can happen pretty easily to provide directory you didn't want to.

Lukas113:
I've implemented the `write_dir` function to Karabo & added it to the telescope `write_to_file` function.

The only thing I'm not really happy about is that firstly it not a write-to-file (since it's a dir) function (however, changing the API is out of scope for this PR I think) and secondly that we don't do any checks if the `overwrite=True` flag is set. I think just checking if it ends with a `.tm` should be sufficient? If you agree, we would have to change some code (each call of `Telescope.write_to_file` would need a .tm ending)

Lukas113:
I've also changed line 475/476 in telescope.py to:
```python
if not os.path.exists(tmp_dir):
    self.write_to_file(tmp_dir)
```
However, we could also do:
```python
self.write_to_file(tmp_dir, overwrite=True)
```
instead to get the `Note` assumption mentioned in the docstring. I think this would be a better choice?

karabo/simulation/telescope.py
anawas: I think there is not much damage in this case if the folder is overwritten. Mainly because this particular funtion will not be used very often. Why should you use it? It's more likely to read a telescope file than to write one. This is because you define the telescope in another software.

anawas (commented):
The naming looks good to me. Now it's clear that data is written into a folder not into a file. 
And enforcing .tm extension ensures that the folder is compatible with OSKAR.

Lukas113 (approved):

Issue #581, opened 2024-06-24
Exception when reading a telescope configuration


anawas:
# Summary
Reading a telescope configuration with `read_OSKAR_tm_file()` can raise an exception. This is true if the layout files `layout.txt` of a station has 6 columns and there is a space at the end of each row, e.g. here:
`1.0, 1.0, 0.0, 0.0, 0.0, 0.0 ` (there is a space at the end of the line)

# Steps to reproduce
Create a telescope, save the configuration to disk, and read it back in:
```
from karabo.simulation.telescope import Telescope
from karabo.simulator_backend import SimulatorBackend

SIMULATOR_BACKEND = SimulatorBackend.OSKAR
telescope = Telescope.constructor("MeerKAT", backend=SIMULATOR_BACKEND)
telescope.write_to_file("mytelescope.tm")

my_tel = Telescope.read_OSKAR_tm_file("mytelescope.tm")
```

# Result
```
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[18], [line 1](vscode-notebook-cell:?execution_count=18&line=1)
----> [1](vscode-notebook-cell:?execution_count=18&line=1) my_tel = Telescope.read_OSKAR_tm_file("mytelescope.tm")

File ~/miniconda3/envs/karabo/lib/python3.9/site-packages/karabo/simulation/telescope.py:561, in Telescope.read_OSKAR_tm_file(cls, path)
    [554]     raise karabo.error.KaraboError(
    [555]         "Could not create Telescope from position.txt file_or_dir. "
    [556]         + "It must contain one line with two or three numbers."
    [557]     )
    [559] position_file.close()
--> [561] station_positions = cls.__read_layout_txt(station_layout_file)
    [562] for station_position in station_positions:
    [563]     telescope.add_station(
    [564]         station_position[0],
    [565]         station_position[1],
   (...)
    [569]         station_position[5],
    [570]     )

File ~/miniconda3/envs/karabo/lib/python3.9/site-packages/karabo/simulation/telescope.py:609, in Telescope.__read_layout_txt(cls, path)
     [607] i = 0
     [608] for pos in station_position:
--> [609]     values[i] = cls.__float_try_parse(pos)
     [610]     i += 1
     [611] positions.append(
     [612]     [values[0], values[1], values[2], values[3], values[4], values[5]]
     [613] )

IndexError: index 6 is out of bounds for axis 0 with size 6
```

# Expected result
[...]

PR #602, opened 2024-07-29
Fix issue #581

anawas:
* Import of OSKAR telescope files with function `read_OSKAR_tm_file` now works.
* I refactored a lot of tests in `test/test_telescope.py`. They used `plot_telescope()` in order test if the Telescope instance was created successfully. This took a lot fo time, especially when only this test was run. The test functions now test other telescope specific values.


Lukas113:
I'll wait for the review until #601 is merged since #601 is also in this PR.

mpluess:
@anawas can you merge main into this branch and solve the conflicts? i will review it after that's done.

mpluess (changes requested):
Looks good. I like that you reduced the test runtime by removing the plot_telescope calls, but i think it would make sense to add two dedicated tests calling plot_telescope to make sure it works, one with OSKAR, one with RASCIL backend since they have different code paths in plot_telescope. The code path with an invalid backend is already covered in the tests.

karabo/simulation/telescope.py
```diff
@@ -402,7 +402,14 @@ def plot_telescope(self, file: Optional[str] = None) -> None:
         if self.backend is SimulatorBackend.OSKAR:
             self.plot_telescope_OSKAR(file)
         elif self.backend is SimulatorBackend.RASCIL:
-            plot_configuration(self.get_backend_specific_information())
+            if file is not None:
```
mpluess, line 405: if-else seems unnecessary, default value of plot_configuration's plot_file is None, so you should be able to just pass plot_file=file in both cases

anawas: Solved in commit 75fe915

mpluess:
Changes look good so far, tests are failing though

anawas:
That's what I talked about on Slack. I did not touch the file the pre-commit hook is complaining about. I ran the tests locally and they passed. This pre-commit hook is bit of hard to deal with.

mpluess:
> That's what I talked about on Slack. I did not touch the file the pre-commit hook is complaining about. I ran the tests locally and they passed. This pre-commit hook is bit of hard to deal with. Screenshot 2024-08-02 115004

I seems hard to have 100 % the same environment as the CI tests on GitHub. I think all of us devs have had to fix things in reaction to tests failing on GitHub before. Had to remove the same "type: ignore" comment in another branch this morning. One difference is probably that on GitHub, a new environment is created on every action, meaning our dev envs are probably a bit behind on same package versions. Updating the environment more often could open.

mpluess (approved):
LGTM

PR #612, opened 2024-08-14
Improve src net rucio script

### LINKED ISSUE #611: Calc coverage from ObsCoreMeta.from_image correctly
Currently, the sky-coverage calculation in `from_image` assumes a flat sky from it's phace-center. However, this leads to incorrect RA/DEC values and can even result in an exception, when the assumed FOV of DEC results in larger than 90 or smaller than -90. This can happen easily for phase-centers near the poles and/or if the FOV is large enough.

To fix this, one needs to calculate the coverage on the sphere, not on a flat approximation using e.g. `WCS.pixel_to_world` & then decide what to do with it (e.g. creating a polygon to approximate coverage).

Lukas113:
closes #611 

According to the SRCNet efforts, a demo to create ObsCore metadata and ingest it into the ingestion daemon is needed. Therefore I adapted the `SRCNet_rucio_meta.py` example script to be a full simulation. This will be helpful for future dataset creation.

In addition, I also discovered a bug by accident, which I described in #611 and is solved in this PR.


karabo/examples/SRCNet_rucio_meta.py
```diff
@@ -1,119 +1,183 @@
...
-    However, we just operate on existing files to avoid example-duplication.
+Here, we create a simulated visibilities and a cleaned image. This is just an example
+script which should be highly adapted, e.g. you custom-sky, simulation params (for
+larger data-products, simulation params as user-input, multiple simulations, etc.
+
+Be aware that API-changes can take place in future Karabo-versions.
+
+If not specified further (e.g. through `XDG_CACHE_HOME` or `FileHandler.root_stm`),
+Karabo is using /tmp. Thus if you have a script which is producing large and/or many
+data products, we suggest to adapt the cache-root to a volume with more space.
```
sfiruch, line 11: Suggestion:

> Here, we simulate visibilities and a cleaned image. This is just an example, which should be adapted for your needs, e.g. provide a custom sky, different simulation parameters, etc.
> 
> If you don't specify anything Karabo will place the data products in the temporary directory `/tmp`. You can change this behavior by setting `FileHandler.root_stm` to a different directory. It could make sense to use a directory on a volume with a lot of free space if you are producing large data sets.

karabo/examples/SRCNet_rucio_meta.py
```diff
@@ -1,119 +1,183 @@
...
+    start_freq_hz = 76e6
+    num_chan = 16
+    freq_inc_hz = 1e8
+
+    if num_chan < 1:
+        err_msg = f"{num_chan=} but must be < 1"
+        raise ValueError(err_msg)
+    if num_chan == 1 and freq_inc_hz != 0:
+        err_msg = f"{freq_inc_hz=} but must be 0 if only one channel is specified"
+        raise ValueError(err_msg)
```
sfiruch, line 52: This should be handled in the `Observation` constructor, not in this example script.

karabo/examples/SRCNet_rucio_meta.py
```diff
@@ -1,119 +1,183 @@
...
+    # specific for your use-case.
+    vis_path = os.path.join(
+        os.path.split(vis.vis_path)[0],
+        f"gleam-ra{phase_center[0]}-dec{phase_center[1]}.vis",
+    )
+    os.rename(
+        vis.vis_path,
+        vis_path,
+    )
+    vis.vis_path = vis_path
```
sfiruch, line 81: Instead of renaming the file after the fact, the right path should be passed to the `InterferometerSimulation` constructor.

karabo/examples/SRCNet_rucio_meta.py
```diff
@@ -1,119 +1,183 @@
...
+    )
+    vis_rm = RucioMeta(
+        namespace="testing",  # needs to be specified by Rucio service
+        name=os.path.split(vis.vis_path)[-1],  # remove path-infos for `name`
+        lifetime=86400,  # 1 day
+        dataset_name=None,
+        meta=vis_ocm,
+    )
+    # ObsCore mandatory fields
+    vis_ocm.obs_collection = "MRO/ASKAP"
```
sfiruch, line 98: Can't this be a parameter of `ObsCoreMeta.from_visibility`?

The same applies to `.obs_id`, and `obs_published_did`

karabo/examples/SRCNet_rucio_meta.py
```diff
@@ -1,119 +1,183 @@
...
+        WscleanImageCleanerConfig(
+            imaging_npixel=imaging_npixel,
+            imaging_cellsize=imaging_cellsize,
+            niter=5000,  # 10 times less than default
         )
-    else:
-        err_msg = f"Unexpected {dp_type=}, allowed are only `dp-type` choices."
-        raise RuntimeError(err_msg)
+    ).create_cleaned_image(  # currently, wsclean needs casa .ms, which is also created
+        ms_file_path=vis.ms_file_path,
```
sfiruch, line 137: We should pass the `vis` object, not file paths.

karabo/examples/SRCNet_rucio_meta.py
```diff
@@ -1,119 +1,183 @@
...
+    # customize fname for restored image for the same reason as visibilities
+    restored_image_path = os.path.join(
+        os.path.split(restored.path)[0],
+        f"gleam-ra{phase_center[0]}-dec{phase_center[1]}.fits",
+    )
+    os.rename(
+        restored.path,
+        restored_image_path,
+    )
+    restored.path = restored_image_path
```
sfiruch, line 150: Instead of renaming after the fact, the imager should generate the image with the desired filename the first time round.

sfiruch (changes requested):
LGTM, but I think the Karabo API should be improved to deal with many details that are laid out in this example.

karabo/examples/SRCNet_rucio_meta.py
Lukas113: As it is currently not really. Firstly, fields like `obs_collection` are not strictly specified and may be specific to a (IVOA Table Access Protocol) TAP service. The same is true for `obs_id` and `obs_published_did`. They can be very specific to a TAP provider  where Karabo has no way of knowing that.

karabo/examples/SRCNet_rucio_meta.py
Lukas113: This might be true, but is out of scope for this PR.

sfiruch: Can you create an issue?

Lukas113: Improve APIs is a general pursue imo. And in this particular case I don't think providing a vis object is an improvement. I think so, because no other information than the ms file-path is needed, which means no vis object utility is needed. And if someone wants to call this function, they would be forced to wrap their MS with a vis object for no reason (if it's not yet a vis object).

It's similar how I design my own API in my other project. If e.g. just the image shape is needed, I usually don't design the function to have to accept the whole image, but just it's shape. It's more general and doesn't force me in any case to artificially create an image when e.g. just shape information is available.

TLDR: I don't think we should create an issue from this.

Lukas113: Another option I could see is to make a positional argument only which could be a ms-path or vis object. However I think just opening an issue because we just have randomly seen this line isn't the way to go. I've also seen other instances of API design I don't like in Karabo. IMO what's needed to improve the API is to generally think about API design and refactor accordingly.

sfiruch (approved):

PR #630, opened 2024-10-31
SRCNet PI 24 run, SKA AA layouts for OSKAR

### LINKED ISSUE #536: Telescope module for AA*
AA* configuration will be used for science benchmarks. We can put this configuration for the science tests.
The expected time line is the quater 4 in 2027 for SKA mid and 2028 for SKA low.

mpluess:
Closes #536


karabo/simulation/telescope.py
```diff
@@ -48,6 +48,16 @@
     CARMAVersions,
     NGVLAVersions,
     PDBIVersions,
+    SKALowAA1Versions,
+    SKALowAA2Versions,
+    SKALowAA4Versions,
+    SKALowAA05Versions,
```
Lukas113, line 54: IMO AA05 for mid and low might be a little misleading. Not sure if sth like 0Dot5 is better? What I mean is not you have to change, but think about it. If you come to the conclusion that the current solution is best, then leave it as it is.

karabo/simulation/telescope_versions.py
```diff
@@ -193,6 +193,46 @@ class PDBIVersions(enum.Enum):
...
+
+
+class SKALowAAStarVersions(enum.Enum):
+    SKA_OST_ARRAY_CONFIG_2_3_1 = "ska-ost-array-config-2.3.1"
+
+
+class SKALowAA4Versions(enum.Enum):
+    SKA_OST_ARRAY_CONFIG_2_3_1 = "ska-ost-array-config-2.3.1"
+
+
```
Lukas113, line 235: Considering how previous telescope-version enums are/were handeled, I wonder why each telescope version has it's own enum here? The previous handling was/is that each telescope has one enum. In our case this would mean one for ska-mid and one for ska-low, and have the according version (e.g. AA2) under this enum.

Lukas113 (changes requested):
Overall great work!

One thing I wonder is how you created the .tm files. With a script? If yes, I think it's work putting that into the repo and don't expose it to public API (somewhere in the data-module, maybe in a private submodule?). This would make future updates easier and improves reproducibility.

mpluess:
> Overall great work!
> 
> One thing I wonder is how you created the .tm files. With a script? If yes, I think it's work putting that into the repo and don't expose it to public API (somewhere in the data-module, maybe in a private submodule?). This would make future updates easier and improves reproducibility.

I had the script and the documentation in the SKA confluence, but I agree that it should rather live in the Karabo repo. I moved it.

karabo/simulation/telescope.py
mpluess: good point, i changed it

karabo/simulation/telescope_versions.py
mpluess: I decided to add each layout as a telescope, e.g. SKA-LOW-AA0.5, and use ska-ost-array-config-[SKA_OST_ARRAY_CONFIG_PACKAGE_VERSION] as the version, e.g. ska-ost-array-config-2.3.1. This was under the assumption that SKA layouts can still have minor changes, which we can then track here using versions. For example, who knows if we'll ever reach AA4 as planned, maybe the final AA4 will have less antennas / stations.

mpluess: I think the pragmatic solution would be to leave it as it is then, since we don't really have a convincing alternative.

mpluess:
Force-merging after addressing review comments due to reviewer being absent for next two weeks.

Issue #629, opened 2024-10-29
Integrate ska-ost-array-config for up-to-date SKA telescope layouts and custom subarrays


mpluess:
PREREQUISITE: Karabo upgrade to Python 3.10

PR https://github.com/i4Ds/Karabo-Pipeline/pull/630 integrates all SKA-Low and -Mid AA layouts into Karabo for use with the OSKAR simulator. See this documentation [Documentation.pdf](https://github.com/user-attachments/files/17557673/Documentation.pdf) and this script [array_config_to_oskar_tm.py.txt](https://github.com/user-attachments/files/17557671/array_config_to_oskar_tm.py.txt) with the accompanying README [README.txt](https://github.com/user-attachments/files/17557674/README.txt).

When the upgrade of Karabo to Python 3.10 is done, ska-ost-array-config should be properly integrated to allow up-to-date SKA telescope layouts and custom subarrays to be directly loaded from there into Karabo. The integration should also allow to use the layouts with RASCIL, not just OSKAR. For RASCIL, we currently only offer the (at least partially outdated) telescopes the come with the RASCIL package.

anawas:
We need to touch the `Telescope` class as well. I ran into an issue this morning. I'm trying to fix it.
This code with **OSKAR** works (i.e. it prints 134 baselines, which is correct):
```
site_tel: Telescope = Telescope.constructor("LOFAR", version=None, backend=SimulatorBackend.OSKAR)
baselines_wgs = site_tel.get_baselines_wgs84()
print(f"{len(baselines_wgs)=}") # len(baselines_wgs)=134
```

The same code with **RASCIL**-backend doesn't work:
```
site_tel = Telescope.constructor("LOFAR", version=None, backend=SimulatorBackend.RASCIL)
baselines_wgs = site_tel.get_baselines_wgs84()
print(f"{len(baselines_wgs)=}")  # len(baselines_wgs)=0
```

The reason is how we initialise the `Telescope` class. It is done differently for instruments defined in OSKAR tm-files and instruments from RASCIL package.

anawas:
Closed in PR #631

mpluess:
Reopening because while some issues were solved in the mentioned PR, ska-ost-array-config is not integrated yet. The idea is to be able to load SKA layouts directly from this library without needing to export them to OSKAR .tm first. Need to wait for the Python 3.10 update though.

anawas:
[...]

PR #631, opened 2024-11-07
Major update of Telescope class

anawas:
Baseline functions of the Telescope class (e.g. `max_baseline` or `get_stations_wgs84`) now have the same output for OSKAR and RASCIL backends. Thus, the two code examples in issue #629 now work as expected. The main work was done in `Telescope::constructor()` function.

Other things that have been changed or adapted:
* Renamed function `get_baselines_wgs84` to `get_stations_wgs84` because it always returned coordinates of stations, not baselines
* Replaced `numpy.loadtxt()` with more robust class method `__read_layout_txt()` (in `Telescope::_get_station_info`)
* Added some tests cases to `test_telescope_baselines.py`
* Fixed some typos and added few lines of comments.


karabo/simulation/sky_model.py
```diff
@@ -825,8 +825,8 @@ def read_from_file(cls: Type[_TSkyModel], path: str) -> _TSkyModel:
 
         if dataframe.shape[1] > cls.SOURCES_COLS:
             print(
-                f"""CSV has {dataframe.shape[1] - cls.SOURCES_COLS + 1}
-            too many rows. The extra rows will be cut off."""
+                f"CSV has {dataframe.shape[1] - cls.SOURCES_COLS + 1} "
+                f"rows too many. The extra rows will be cut off."
```
Lukas113, line 829: line 829 doesn't need to be an f-string.

karabo/simulation/coordinate_helper.py
```diff
@@ -19,10 +19,10 @@ def east_north_to_long_lat(
     """
 
     # https://stackoverflow.com/questions/7477003/calculating-new-longitude-latitude-from-old-n-meters
-    r_earth = 6371000
-    new_latitude = lat + (east_relative / r_earth) * (180 / np.pi)
-    new_longitude = long + (north_relative / r_earth) * (180 / np.pi) / np.cos(
-        long * np.pi / 180
+    r_earth = 6378100  # from astropy (astropy.constants.R_earth.value)
```
Lukas113, line 22: nice!

karabo/simulation/telescope.py
```diff
@@ -277,27 +279,67 @@ def constructor(
...
+    def __convert_to_karabo_telescope(cls, instr_name: str) -> Telescope:
+        """Converts a site saved in RASCIl data format into a Karabo Telescope.
+
+        :param instr_name: The name of the instrument to convert.
+
+        :returns: An instance of Karabo Telescope.
+        :rtype: karabo.simulation.telescope.Telecope
+        :raises: ValueError if instr_name is not a valid RASCIL telescope
+        """
+        config: Configuration = create_named_configuration(instr_name)
```
Lukas113, line 316: does specifying the dtype `Configuation` do anything? If not (I assume because the function should return just this type), I think removing would be a better idea. Not sure what the implication of this is in case the function is changing it's return type for each use-case.

karabo/simulation/telescope.py
```diff
@@ -277,27 +279,67 @@ def constructor(
...
+        for i, coord in enumerate(station_coords):
+            telescope.add_station(
+                horizontal_x=coord[0],
+                horizontal_y=coord[1],
+                horizontal_z=coord[2],
+            )
+
+            # there are only stations in the rascil files no antennas
+            # we add a dummy antenna
+            telescope.add_antenna_to_station(i, 0.1, 0.1)
```
Lukas113, line 339: I don't understand why both horizontal x & y need to be 0.1. Can you elaborate?

karabo/simulation/telescope.py
```diff
@@ -485,7 +536,7 @@ def write_to_disk(self, dir_name: DirPathType, *, overwrite: bool = False) -> No
         ):  # for OSKAR & `overwrite` security purpose
             err_msg = f"{dir_name=} has to end with a `.tm`, but doesn't."
             raise RuntimeError(err_msg)
-        with write_dir(dir=dir_name, overwrite=overwrite) as wd:
+        with write_dir(dir=dir_name, overwrite=True) as wd:
```
Lukas113, line 539: I think this is an accident? If not, what's the motivation behind ignoring the fun-arg?

karabo/test/test_rascil_telescope_setup.py
```diff
@@ -0,0 +1,50 @@
...
+    site: Telescope = Telescope.constructor(site_name, backend=SimulatorBackend.RASCIL)
+    assert len(site.stations) == num_stations
+
+
+# @pytest.mark.parametrize("site_name, num_stations", rascil_telesecopes_to_test)
+# def test_num_of_baselines(site_name, num_stations):
+#         site: Telescope =
+# Telescope.constructor(site_name, backend=SimulatorBackend.RASCIL)
+#         num_baselines = num_stations * (num_stations-1) // 2
+#         assert len(site.get_baselines_wgs84()) == num_baselines
```
Lukas113, line 50: Why out-commented? If not needed, I think just removing it would be better?

karabo/test/test_telescope_baselines.py
```diff
@@ -67,5 +98,47 @@ def test_baselines_based_cutoff(sky_data: NDArray[np.float64]):
...
+def test_telescope_stations(oskar_telescope: Telescope, rascil_telescope: Telescope):
+    # station has 30 stations according to *.tm file
+    baseline_wgs = oskar_telescope.get_stations_wgs84()
+    assert len(baseline_wgs) == 64
+
+    baseline_wgs = rascil_telescope.get_stations_wgs84()
+    assert len(baseline_wgs) == 134
+
+
+def test_telescope_baseline_length(rascil_telescope):
```
Lukas113, line 140: I think adding type-hint here would be an improvement. In case someone does a refactoring on `get_stations_wgs84`, it would be catched automatically. Otherwise you'll just see the fail in the failing CI-pipeline.

Lukas113 (changes requested):
LGTM

Some comments to address.

The failing CI-pipeline is not related to this PR, thus the test is ignored for this review.

karabo/simulation/sky_model.py
anawas: Will be fixed. And also on the lines 822, 823 above.

karabo/simulation/telescope.py
anawas: Sorry, I am not sure what you mean. The function `create_named_configuration` is from the `ska_sdp_datamodels` package. It is used to parse the RASCIL config files properly. I need to call it. Otherwise I need to take care of reading theses files myself.

karabo/simulation/telescope.py
anawas: Not setting a dummy antenna creates an empty file. I think that's not a goot idea. And it matches most of the telescope files in `karabo/data`. See config file `karabo/data/aca.all.tm/station000/layout.txt` for example.

karabo/simulation/telescope.py
anawas: Don't know what I did here. Reverted it to original code.

karabo/simulation/telescope.py
Lukas113: My comment is about `config: Configuration = create_named_configuration(instr_name)` vs `config = create_named_configuration(instr_name)`

karabo/simulation/telescope.py
Lukas113: Hmm, ok. I now had a closer look into `telescope.py` and I have some feedback/questions (not only from your code, also from previous authors):

- I see for dishes (not antenna clusters) there is a 0.1, 0.1 in other telescopes as well. However, I don't get why this is not a 0.0 instead because it's the position relative to the station center. And in my understanding, the station and dish center should be the same? Can you elaborate on this?
- To me it seems like this function only works for dish-arrays, not for antenna-arrays because of the hard-coded `telescope.add_antenna_to_station`. However, looking at possible values from `create_named_configuration` this assumption doesn't hold. I suggest to change the function that it works for both telescope-types.



NOT from your PR, but low-hanging fruit improvements:

- Please add `raise IndexError(err_msg)` to an else part in `add_antenna_to_station`. Otherwise it can cause a silent error which is not a good design choice.
- plot stations only, not also antennas in `plot_telescope_OSKAR`. They should/must have similar coordinates either way and just "spam" the plot with dots on the same location.

anawas: > plot stations only, not also antennas in plot_telescope_OSKAR. They should/must have similar coordinates either way and just "spam" the plot with dots on the same location.

I suggest to give it a try. For SKA-Mid each station has one antenna (dish). But there are 128 antennas for each station for SKA-Low. But the anetnnas are barely visibile on the scale of the image.

karabo/simulation/telescope.py
anawas: Type hint removed. The return type can be deduced from the function definition.

karabo/test/test_telescope_baselines.py
anawas: Donr

anawas:
I would like to close this pull request. There is more work to do for the `Telescope` class (see issue #633). And I would like to address the open comment regarding the `telescope.add_antenna_to_station` function there.

Lukas113:
> I would like to close this pull request. I addressed most of the comments. But there is more work to do for the `Telescope` class (see issue #633). Tackling these issues here would clearly overload this PR. Thus I suggest to address the open comment regarding the `telescope.add_antenna_to_station` function there.

What's the reason why you want this to be merged asap? Because of the release? I personally don't like to introduce potential misbehavior into main, when the issue should get addressed anyway. I am more in favor of leaving this PR open and you can focus on the other issue in a separate branch in case you want to merge it before the release.

PR #634, opened 2024-12-04
Added latest ALMA cycles configurations

anawas:
* In the upcoming year 2025 the ALMA telescope (Atacama Large Microwave Array) offers three new dish configurations (called cycles). I added configuration files for these.
* I also added the script which I used to convert the ALMA config files into OSKAR telescope configurations. It can be used both to convert further cycles, and update past cycles.


karabo/data/_add_oskar_alma_layouts/README.md
```diff
@@ -0,0 +1,12 @@
+## What this script is for
+The dishes of the ALMA telescope (Atacama Large Microwave Array) can operate in different configurations. These 'cycles' are set up at different points in time. For the next year (2025) three new cycles are added to the configuration: cycle 9, 10 and 11. This script can be used to download the latest or update the current configuration files.
+
+The files are fetched directly from the ALMA server. The url is not under our control. Therefore, the url may change which breaks the code. In this case have a look at link [2] and update the variable `CONFIG_FILE_URL` in the code.
+
+##  Setup
+1. Create and activate a Python 3.10 (3.11 should also work, not tested) virtual env / conda env / similar
```
mpluess, line 7: Is Python 3.10 really a requirement or is that just a remnant from my SKA layouts script? The Python 3.10 requirement there stems from the ska_ost_array_config library, which you don't use here.

karabo/data/_add_oskar_alma_layouts/README.md
```diff
@@ -0,0 +1,12 @@
+## What this script is for
```
mpluess, line 1: I like that you put the documentation right next to the script. I had put it to the dev docs here: https://i4ds.github.io/Karabo-Pipeline/development.html#add-ska-layouts-as-oskar-telescopes. I still think it would make sense to add a very brief paragraph similar to mine to the dev docs just to increase the chances that if someone else wants to add ALMA layouts in the future, they find your existing code.

mpluess (changes requested):
Two suggestions, looks good otherwise!

karabo/data/_add_oskar_alma_layouts/README.md
anawas: Ok, fair enough. Thus, there is no need to restrict the version. I tested it with Python 3.12 and it works.
I'll remove the version hint.

karabo/data/_add_oskar_alma_layouts/README.md
anawas: I was not aware of this page. So I'll add my documentation there as well.

mpluess (approved):
LGTM

PR #639, opened 2025-01-23
Updating Documentation

anawas:
As we seem to be working towards a Karabo version 1.0 I'm about to update the documentation.
What I did so far:
* Add documentation to many modules in _karabo.util_ package
* Rewriting of many docstring to the Google format.

### DIFF
--- a/doc/src/development.md
+++ b/doc/src/development.md
@@ -73,7 +73,7 @@ If you run mypy several times for the same file(s) and it takes a while, you can
 
 ### pydocstyle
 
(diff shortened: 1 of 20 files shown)


Lukas113 (approved):
LGTM

Two mentions:

[Pydocstyle](https://github.com/PyCQA/pydocstyle) is deprecated, [Ruff](https://github.com/astral-sh/ruff) should be used instead (since you made changes in the pydocstyle doc). It shouldn't matter in the short-term I think? But we need to change that at some point. It should also be integrated into the CI/CD pipeline & dev-dependencies. But since the repo isn't ready to be tested in the by a docsting-tool ATM I don't think it's critical and needs to be changes in this PR.

Secondly, type-hints in docstrings are not needed if the signatures are typed (which should be the case in our repo). But I leave it up to you whether you wanna change that or not.

Issue #640, opened 2025-02-12
Publication associated with this software?


tmartinezML:
Hello dear Contributors,

I am writing a paper on simulations of radio observations, and I would like to mention your software in the introduction.
I was wondering if there is a scientific paper associated with this software which I might cite? Also, do you happen to know of any other publications where this software was used in their work?
Thanks!

anawas:
Solved with PR #641

PR #641, opened 2025-02-13
Added citation file

anawas:
I added a citation file as described here: 
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files

This commit resolves issue #640.

### DIFF
--- a/.github/workflows/update-citation-cff.yml
+++ b/.github/workflows/update-citation-cff.yml
@@ -1,0 +1,43 @@
+name: Update CITATION.cff
+on:
+  release:
+    types: released
+    branches:
+      - main 
+
+jobs:
+  update-citation-cff:
+    runs-on: ubuntu-latest
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: main
+          fetch-depth: 0
+
+      - name: Update CITATION.cff
+        run: |
+          # Update CITATION.cff file
+          git fetch --tags
+          latest_tag=$(git describe --tags $(git rev-list --tags --max-count=1))
+          echo "The latest published version is $latest_tag"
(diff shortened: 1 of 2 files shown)


Lukas113 (commented):
Generally LGTM

However, I'm not sure about two things:

- Firstly about the authors. Some contributors are included and others are not (e.g. you Andreas, Luis, Philip, Jennifer and Christoph). However, I'm not exactly what to do about this other than include them all? IDK.
- Secondly about the version and release date. These change with a new release and have to be updated each time, which might easily get forgotten. Are they mandatory?

anawas:
I suggest to use the affiliation instead of a list of authors. What I can do with the cff file is:

FHNW, School of Computer Science, Institute for Data Science. (2024). Karabo (Version 0.31.0) [Computer software]. https://github.com/i4Ds/Karabo-Pipeline

The version of a software is essential for the traceability of scientific results. But I see your point that one will forget to bump it in this file. Since the cff file is a basic text file, can we generate it from a template every time we create a new release?

Lukas113:
:+1: for affiliation

ok. The best solution would be to update the version automatically in the CI from a release-event. I think this could be achieved by  [this workflow](https://github.com/sourmash-bio/sourmash/issues/3080) which does the update. However, this would just be a commit after the release. The tag would need to be moved from last-commit to current HEAD afterwards with sth like [this](https://www.geeksforgeeks.org/how-to-move-tag-on-a-git-branch-to-different-commit/).

So I suggest if you have time and are interested to do some CI/CD work go for it. But I understand that this is some effort to just get a little automation gain. Otherwise I suggest do adapt the `doc.src.development.md` "Create a Release" section to adjust the version right before a release is triggered.

anawas:
#### What changed:

- The file CITIATION.cff now lists affiliation instead of individuals
- I added an action workflow that updates the version in this file when a new release is published. I followed the links in the comment by @Lukas113.

Lukas113 (approved):
LGTM

:crossed_fingers: that it works :)

PR #644, opened 2025-03-25
Workflow execution script

Lukas113:
- Adds SRCNet workflow execution file to examples with env-var configurability (see Jira CHOC-185).
- Adds `Environment` class to enable easy env-var parsing.

### DIFF
--- a/karabo/examples/SRCNet_workflow_AAstar.py
+++ b/karabo/examples/SRCNet_workflow_AAstar.py
@@ -12,6 +12,7 @@
 import math
 import os
 from datetime import datetime, timedelta
+from typing import cast
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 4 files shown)


Lukas113:
Could be WiP depending on what we discuss in SRCNet. Therefore review is welcome, but shouldn't be merged just yet.

karabo/examples/SRCNet_workflow_AAstar.py
```diff
@@ -0,0 +1,284 @@
...
+)
+
+# Imaging
+# Image size in degrees should be smaller than FOV
+# Bigger baseline -> higher resolution
+# Image resolution from SKAO's generate_visibilities.ipynb
+IMAGING_NPIXEL = Environment.get("IMAGING_NPIXEL", int, 20000)
+# -> Cellsize < FOV / 20000 -> 9.32190458333e-7
+IMAGING_CELLSIZE = Environment.get("IMAGING_CELLSIZE", float, 9.3e-7)
+
```
anawas, line 67: Should we assert that cellsize < FOV / 20000?

karabo/examples/SRCNet_workflow_AAstar.py
```diff
@@ -0,0 +1,284 @@
...
+from karabo.simulator_backend import SimulatorBackend
+from karabo.util.gpu_util import is_cuda_available
+from karabo.util.helpers import Environment
+
+# Simulation
+# Phase center: Must be inside coverage region of selected sky
+SKY_MODEL = Environment.get("SKY_MODEL", str, "MIGHTEE_L1")
+PHASE_CENTER_RA_DEG = Environment.get("PHASE_CENTER_RA_DEG", float, 150.12)
+PHASE_CENTER_DEC_DEG = Environment.get("PHASE_CENTER_DEC_DEG", float, 2.21)
+START_FREQ_HZ = Environment.get("START_FREQ_HZ", float, 1.304e9)
```
anawas, line 44: Possible issue here: Is there a check down the line that the phase center coordinates are within the sky model? If they're not the simulation will run for hours but will not generate meaningful output. It could be user friendly to 'fail fast' here.

anawas (approved):
Can be merged. The question if we should use the OSKAR backend in this frequency range has been submitted. The answer is still pending.

anawas (approved):
LGTM. Adding a comment about the cell size used provides valuable information to the user in this case.

PR #645, opened 2025-04-01
SP-5327/CHOC-232 MWA ObsCore metadata improvements

d3v-null:
This is a draft to show my progress towards SP-5327 - Improve discoverability of MWA data via TAP/Rucio.

So far I've only completed part of choc-232:
- fixed angular resolution calculation for s_resolution
- swap em_min and em_max
- fix calculation of t_xel, number of timesteps

at this point the tests pass, but I noticed an issue with one of the test datasets, and will check with the team what to do about it.

after that I intend to extend the tests to MWA flavoured measurement sets, uvfits, uvh5.

### DIFF
--- a/karabo/data/obscore.py
+++ b/karabo/data/obscore.py
@@ -542,11 +542,11 @@ class ObsCoreMeta:
                     nrow=300000000,  # to avoid memory & time issues, should be repres
                 )
             )
-            ocm.t_xel = len(np.unique(MSMainTable.get_col(
-                ms_path=vis_inode,
-                col="TIME",
-                nrow=300000000
-            )))
+            ocm.t_xel = len(
+                np.unique(
+                    MSMainTable.get_col(ms_path=vis_inode, col="TIME", nrow=300000000)
+                )
+            )
             spectral_window = ms_meta.spectral_window
             # wavelength = c/f, min is the highest frequency, max is the lowest
             ocm.em_max = c / np.min(
--- a/karabo/data/obscore.py
+++ b/karabo/data/obscore.py
@@ -495,6 +495,10 @@ class ObsCoreMeta:
 
         Returns:
             `ObsCoreMeta` instance.
+
+        Updated 2025-04-01:
+            - em_min is now less than em_max
+            - t_xel is now the number of unique time-steps, not the number of rows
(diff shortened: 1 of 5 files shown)


karabo/data/obscore.py
```diff
@@ -495,6 +495,10 @@ def from_visibility(
 
         Returns:
             `ObsCoreMeta` instance.
+
+        Updated 2025-04-01:
+            - em_min is now less than em_max
+            - t_xel is now the number of unique time-steps, not the number of rows
```
Lukas113, line 501: As far as I know, version update & bugfix comments don't belong in a docstring. Instead in an inline comment or just in the commit-msg.

Lukas113 (changes requested):
LGTM, thanks!

There's only one comment about docstring to address.

d3v-null:
Thanks for the review @Lukas113 

Let me know when you've had a chance to update the `minimal_casa_ms` fixture. Maybe you already have, but I'm not sure how to clear my cache.

d3v-null:
Thanks @Lukas113 , If you're happy with this PR, then you can go ahead and merge. I don't think there's much point in waiting for my other issue.

Lukas113 (dismissed):
LGTM, thanks!

Lukas113:
@d3v-null unfortunately [CI-pipeline](https://github.com/i4Ds/Karabo-Pipeline/blob/42a2d5d23403d032f5b0204a3bca9056d3a8070d/.github/workflows/test.yml#L35) didn't pass. Be sure to follow the coding-requirements to pass the CI by just running the same commands locally. I personally like to work with pre-commit-hooks to avoid some of the issues. It basically prevents you to do a commit which didn't pass your hooks requirement. In case you want to use it, you first have to install  (Karabo installs it with `pip install -e ".[dev]"`) and enable it in the repo root `pre-commit install`.

d3v-null:
Thanks @Lukas113 , I'll make sure to do that next time. 
Is there a way that the CI could run automatically for my PRs? 
image

d3v-null:
also this may be a local issue, but there's one linter warning remaining.
image

Lukas113 (approved):

PR #646, opened 2025-04-29
[choc-233,234] add MWA uvfits obscore support via pyuvdata

d3v-null:
- add pyuvdata 2.4.1 as a dependency. currently pinned because for python 3.9 compatibility
- use pyuvdata for ms and uvfits reading
- add small test datasets for mwa ms and uvfits, uploaded to CSCS as test fixtures
- use future array shapes in pyuvdata
- auto-format fixes

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,7 +29,7 @@ jobs:
         run: |
           conda create -y -n test_karabo python=3.9
           conda activate test_karabo
-          mamba env update --file environment.yaml
+          conda env update -f environment.yaml
           pip install -e ".[dev]"
           python -m ipykernel install --user --name python3
       - name: Test Dev-Tools
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -66,8 +66,13 @@ requirements:
     # exclude buggy versions of other tools
     - setuptools              !=71.0.0, !=71.0.1, !=71.0.2  # buggy with `importlib_metadata`
     # for MWA measurement set, uvfits, hdf5, mir support (choc-234)
-    # pinnned to 2.4.1 because later versions use an astropy that needs numpy 2
-    - pyuvdata[casa]          <=2.4.1
+    # pinnned to 2.4.1 because:
+    # - 2.4.2 has a corrupt file in conda-forge
+    # - later versions use an astropy that needs numpy 2
+    - pyuvdata                <=2.4.1
+    # the following enables casacore in python. specifying the above dep as pyuvdata[casa] works in
+    # conda but mamba doesn't understand optional dep syntax
+    - python-casacore         <=3.7.1
 
 test:
   imports:
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -66,13 +66,8 @@ requirements:
     # exclude buggy versions of other tools
     - setuptools              !=71.0.0, !=71.0.1, !=71.0.2  # buggy with `importlib_metadata`
     # for MWA measurement set, uvfits, hdf5, mir support (choc-234)
-    # pinnned to 2.4.1 because:
-    # - 2.4.2 has a corrupt file in conda-forge
-    # - later versions use an astropy that needs numpy 2
-    - pyuvdata                <=2.4.1
-    # the following enables casacore in python. specifying the above dep as pyuvdata[casa] works in
-    # conda but mamba doesn't understand optional dep syntax
-    - python-casacore         <=3.7.1
+    # this [casa] syntax won't work with mamba!
+    - pyuvdata[casa]          >=2.4,<3
 
 test:
   imports:
--- a/conda/meta.yaml
+++ b/conda/meta.yaml
@@ -65,6 +65,9 @@ requirements:
     - conda-forge::fftw       =*=mpi_mpich*
     # exclude buggy versions of other tools
     - setuptools              !=71.0.0, !=71.0.1, !=71.0.2  # buggy with `importlib_metadata`
+    # for MWA measurement set, uvfits, hdf5, mir support (choc-234)
+    # pinnned to 2.4.1 because later versions use an astropy that needs numpy 2
+    - pyuvdata[casa]          <=2.4.1
 
 test:
   imports:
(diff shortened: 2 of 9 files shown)


Lukas113:
@d3v-null isn't it possible to use pyuvdata >= 3.1.3? Because if not, I'm not sure how to integrate it into Karabo properly since that's the lowest version available for [pyuvdata conda-forge](https://anaconda.org/conda-forge/pyuvdata/files).

d3v-null:
Hmmmmm, it seems like the issue is that mamba handles the package specification differently to conda and pip.

I can reproduce the issue by creating a fresh conda environment

```bash
mkdir mamba_test
cd $_
cat << EOF > environment.yaml
channels:
  - i4ds
  - nvidia/label/cuda-11.7.0
  - conda-forge
dependencies:
  - python                  =3.9
  - pyuvdata[casa]          <=2.4.1
EOF
```

In the Dockerfile, the environment is built with `conda env update -f="environment.yaml"` , and this works fine.

```bash
conda create -y -n test_conda python=3.9
conda activate test_conda
conda env update -f="environment.yaml"
python -c 'import pyuvdata; uvd=pyuvdata.UVData.from_file("test.ms"); print(uvd.data_array.shape)'
```

On the other hand, setting up the environment with mamba using `mamba env update --file environment.yaml` fails

```bash
conda deactivate
conda create -y -n test_mamba python=3.9
conda activate test_mamba
conda install -y -c conda-forge mamba
mamba env update --file environment.yaml

error    libmamba Could not solve for environment specs
    The following package could not be installed
    └─ pyuvdata[casa] <=2.4.1 * does not exist (perhaps a typo or a missing channel).
critical libmamba Could not solve for environment specs
```

I think it's because of the optional dependency syntax in `pyuvdata[casa]`, which conda and pip understand but not mamba.

Why do we use mamba in the tests but conda in the Dockerfile?

Either way, changing the environment.yaml to the following seems to work for both conda and mamba

```yaml
cat << EOF > environment2.yaml
channels:
  - i4ds
  - nvidia/label/cuda-11.7.0
  - conda-forge
dependencies:
  - python                  =3.9
  - pyuvdata                <=2.4.1
  - python-casacore         <=3.7.1
EOF
mamba env update --file environment2.yaml
python -c 'import pyuvdata; uvd=pyuvdata.UVData.from_file("test.ms"); print(uvd.data_array.shape)'
```

<details>

<summary>Additional info: pyuvdata 3.1.3 definitely requires python 3.10, and pyuvdata 2.4.1 is definitely available on conda forge</summary>

```
conda install -c conda-forge 'pyuvdata[casa]<=3.1.3'
The following packages are incompatible
├─ pin-1 is installable and it requires
│  └─ python 3.9.* , which can be installed;
└─ pyuvdata 3.1.3  is not installable because there are no viable options
   ├─ pyuvdata 3.1.3 would require
   │  └─ python >=3.10,<3.11.0a0 , which conflicts with any installable versions previously reported;
```

installing `pyuvdata[casa]<=3.1.3` just uses pyuvdata 2.4.2 which is on numpy 1 ( i thought it said this version wasn't available )

```
conda install -c conda-forge 'pyuvdata[casa]<=3.1.3'
The following NEW packages will be INSTALLED:
astropy            conda-forge/linux-64::astropy-6.0.1-py39h44dd56e_0 
numpy              conda-forge/linux-64::numpy-1.26.4-py39h474f0d3_0
pyuvdata           conda-forge/linux-64::pyuvdata-2.4.2-py39hddac248_0 
```

`pyuvdata[casa]==2.4.1` is also definitely available on conda-forge.

```
conda install -c conda-forge 'pyuvdata[casa]==2.4.1'
  numpy              conda-forge/linux-64::numpy-1.26.4-py39h474f0d3_0 
  pyuvdata           conda-forge/linux-64::pyuvdata-2.4.1-py39hddac248_1 
```

</details>

d3v-null:
I'm noticing that the other notebooks also get skipped in CI. Can I do the same for imaging.ipynb if there's no obvious solution?

```
@pytest.mark.skipif(IS_GITHUB_RUNNER or not RUN_NOTEBOOK_TESTS ...
```

d3v-null:
I barely changed anything and the oskar test passed again, which suggests that the oskar issue in imaging.ipynb is temperamental

d3v-null:
also the codecov upload failed because of a rate limit.

```
There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 429 - {"message":"Rate limit reached. Please upload with the Codecov repository upload token to resolve issue. Expected time to availability: 2146s."}
```

Lukas113:
- `Can I do the same for imaging.ipynb if there's no obvious solution?`: As I see, it should already be the case (see [code](https://github.com/i4Ds/Karabo-Pipeline/blob/71252c758b192e8098b8b00f628e46709bf37076/karabo/test/test_notebooks.py#L78))

- You're right about pyuvdata versions on conda-forge. For some reason I just didn't find the `next` button XD. However, you described that you restrict the version because of Python version. I don't think you should care about  that, because it restricts itself through the Python version. IMO a more reasonable reason for upper-bounding would be if newer versions have API breaking changes from code used in Karabo. Otherweise I think it's better to leave the version open. The same goes for casacore.

- For `conda env update ...` vs `mamba env update ...`. I think it's a historical thing? Because since [Conda 23.10](https://github.com/conda/conda/releases/tag/23.10.0) libmamba solver is default. And therefore I think having `conda env update ...` everywhere is the right thing to do. So I think the "best" way to do it is to replace the `mamba` with `conda`, and do the [casa] thing? Otherwise there would be a "hidden" dependency of pyuvdata and casacore (even tough we have a direct dependency).

The rest LGTM

d3v-null:
Thanks for the review @Lukas113 

1. Some of the notebooks skip if `IS_GITHUB_RUNNER or not RUN_NOTEBOOK_TESTS`, but the imaging notebook only skips if `not RUN_NOTEBOOK_TESTS`. Adding a skip condition if `IS_GITHUB_RUNNER` would stop the flaky test from running in CI. 

2. I like the idea of letting the package impose its own version restrictions, and have conda handle those for us. So shall we have the restriction be `>=2.4,<3.0`, assuming the jump from pyuvdata 2.4 to 3.0 is a breaking change? I'll try it out

3. Having the test environment set up the same way as the development environment makes sense, so I'll update the test CI from mamba to conda

Lukas113 (approved):
LGTM :+1:

PR #653, opened 2025-07-21
Upgrade Python to 3.10

d3v-null:
This is a rebase of Delberin's `upgrade_python` branch, but without the CUDA upgrade which can be done separately.

changes:
- CI: added python version parameter to build-docker-image, conda-build, test-user-package
- CI: tests workflow uses python3.10
- Dockerfile takes python version parameter
- Conda build config: add python3.10, pin numpy 1.26.4
- Conda env: upper bound for scipy

Most of the CI passes:
- Tests (which now uses python 3.10): https://github.com/i4Ds/Karabo-Pipeline/actions/runs/16335361958
- Conda Build which builds wheel for python 3.9 and 3.10, and Dockerfile for python 3.10 : https://github.com/i4Ds/Karabo-Pipeline/actions/runs/16335354060 
- Test User Package CI has never been run before, is not working, probably unrelated to these changes. https://github.com/i4Ds/Karabo-Pipeline/actions/runs/16337026833


Dockerfile
```diff
@@ -1,18 +1,23 @@
 FROM nvidia/cuda:11.7.1-cudnn8-devel-ubuntu22.04
 # build: user|test, KARABO_VERSION: version to install from anaconda.org in case build=user: `{major}.{minor}.{patch}` (no leading 'v')
-ARG GIT_REV="main" BUILD="user" KARABO_VERSION=""
+ARG GIT_REV="main"
+ARG BUILD="user"
+ARG KARABO_VERSION=""
+ARG PYTHON_VERSION="3.10"
```
Lukas113, line 6: I personally like it to have just one docker build step for each env-var, not multiple. afaik each docker command in the Dockerfile adds an additional layer.

However, feel free to agree or disagree on this opinion.

Dockerfile
```diff
@@ -1,18 +1,23 @@
...
+
 RUN apt-get update && apt-get install -y git gcc gfortran libarchive13 wget curl nano
 ENV LD_LIBRARY_PATH="/usr/local/cuda/compat:/usr/local/cuda/lib64" \
     PATH="/opt/conda/bin:${PATH}" \
     IS_DOCKER_CONTAINER="true"
-RUN wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-py39_23.5.0-3-Linux-x86_64.sh -O ~/miniconda.sh && \
+RUN export PY_VERSION_SHORT=$(echo ${PYTHON_VERSION} | sed 's/\.//g') && \
+    wget --quiet https://repo.anaconda.com/miniconda/Miniconda3-py${PY_VERSION_SHORT}_23.5.0-3-Linux-x86_64.sh -O ~/miniconda.sh && \
```
Lukas113, line 13: I don't see the need to put in the python version for according conda installation, because it just affects the base environment. And below we're using a different venv than base.

conda/conda_build_config.yaml
```diff
@@ -1,2 +1,6 @@
 python:
   - 3.9
+  - 3.10
+
+numpy:
+  - 1.26.4
```
Lukas113, line 6: I don't like to pin numpy version at all. This has a lot of implications on other dependencies and may introduce additional technical debt.

Why is this necessary over lower-upper bounding numpy?

conda/meta.yaml
```diff
@@ -54,11 +54,11 @@ requirements:
     - reproject               >=0.9,<=10.0
     - requests
     - rfc3986                 >=2.0.0
-    - scipy                   >=1.10.1
+    - scipy                   >=1.10.1,<1.14 # upper-bounded because healpy 1.15.0 import scipy.integrate.trapz
```
Lukas113, line 57: Is this a Karabo restriction? AFAIK it's from a dependency.

If it is from a dependency and not from Karabo, it shouldn't be here IMO.

environment.yaml
```diff
@@ -40,11 +40,11 @@ dependencies:
   - reproject               >=0.9,<=10.0
   - requests
   - rfc3986                 >=2.0.0
-  - scipy                   >=1.10.1
+  - scipy                   >=1.10.1,<1.14 # upper-bounded because healpy 1.15.0 import scipy.integrate.trapz
```
Lukas113, line 43: same comment as in meta.yaml

Lukas113 (changes requested):
mostly LGTM

please comment, resolve or address according review comments.

conda/conda_build_config.yaml
Delberin-Ali: Numpy wasn't installed in the final Docker environment, likely due to a solver edge case where it wasn't pulled in by any other dependency.
To resolve the ModuleNotFoundError, I pinned numpy=1.26.4 explicitly so it's guaranteed to be present.

This is why I pinned that( last Month). Maybe @d3v-null has something to say about this.

conda/meta.yaml
Lukas113: can you point me to that section? I haven't found it myself.

conda/conda_build_config.yaml
Lukas113: In my head, it doesn't make sense how this measure can resolve the issue mentioned. And because it's introducing additional technical debt, it's not something I just want to accept.

conda/meta.yaml
d3v-null: it's not becuase karabo directly calls this, but because a few packages (bipp, tools21cm) call `from scipy.integrate import trapz` 

even though each of these packages have had their scipy versions constrained, conda seems to ignore this, and build with the wrong scipy version anyway if you don't put this constraint explicitly in karabo.

d3v-null: oops, after double checking this, I just noticed the issue is in tools21cm

```
scipy >=1.10.1,<1.15
```

it should be 

```
scipy >=1.10.1,<1.14
```

instead.

d3v-null: oh no wait, 

tools21cm needs scipy <1.15 because scipy.integrate.quadrature is deprecated in 1.15 https://github.com/i4Ds/Karabo-Feedstock/issues/30

but bipp needs scipy <1.14 because scipy.integrate.trapz is deprecated in 1.14 

either way, both packages have the correct version constraints, but from my understanding, things break when you remove the version constraint from karabo. I'll see if it works if I remove this version constraint from karabo.

Lukas113: I would expect that it works if the constraint is just in tools21cm (as it should be)

conda/conda_build_config.yaml
d3v-null: adding `conda install "conda-build>=25"` caused the build to time out after 6 hours here: https://github.com/i4Ds/Karabo-Pipeline/actions/runs/16165994343/job/45627921591

conda/meta.yaml
Lukas113: Knowing that, I thing we might have to put this into Karabo if that's the only way how it works.... But with a comment why it's there! Not just the error-msg, but because of the restriction of bluebild which is not working.

Dockerfile
Lukas113: With the base env, I think we can go as high as possible. I don't see a reasony why to stick with older versions

d3v-null: I agree, but here's what happens when I switch to the latest available miniconda, `https://repo.anaconda.com/miniconda/Miniconda3-py313_25.5.1-1-Linux-x86_64.sh`

They just HAD to add some bullshit about accepting terms and services 🙄️

```
 => ERROR [4/9] RUN conda install -y -n base conda-libmamba-solver &&     conda config --set solver libmamba &&     conda create -y  3.3s 
------                                                                                                                                    
 > [4/9] RUN conda install -y -n base conda-libmamba-solver &&     conda config --set solver libmamba &&     conda create -y -n karabo python=3.10:                                                                                                                                 
3.049                                                                                                                                     
3.049 CondaToSNonInteractiveError: Terms of Service have not been accepted for the following channels. Please accept or remove them before proceeding:                                                                                                                              
3.049     • https://repo.anaconda.com/pkgs/main
3.049     • https://repo.anaconda.com/pkgs/r
3.049 
3.049 To accept a channel's Terms of Service, run the following and replace `CHANNEL` with the channel name/URL:
3.049     ‣ conda tos accept --override-channels --channel CHANNEL
3.049 
3.049 To remove channels with rejected Terms of Service, run the following and replace `CHANNEL` with the channel name/URL:
3.049     ‣ conda config --remove channels CHANNEL
3.049 
3.049 
3.049 ERROR conda.cli.main_run:execute(127): `conda run /bin/bash -c conda install -y -n base conda-libmamba-solver &&     conda config --set solver libmamba &&     conda create -y -n karabo python=${PYTHON_VERSION}` failed. (See above for error)
------
Dockerfile:13
--------------------
  12 |     SHELL ["conda", "run", "-n", "base", "/bin/bash", "-c"]
  13 | >>> RUN conda install -y -n base conda-libmamba-solver && \
  14 | >>>     conda config --set solver libmamba && \
  15 | >>>     conda create -y -n karabo python=${PYTHON_VERSION}
  16 |     # change venv because libmamba solver lives in base and any serious environment update could f*** up the linked deps like `libarchive.so`
--------------------
ERROR: failed to solve: process "conda run -n base /bin/bash -c conda install -y -n base conda-libmamba-solver &&     conda config --set solver libmamba &&     conda create -y -n karabo python=${PYTHON_VERSION}" did not complete successfully: exit code: 1
```

so I'm going with updating to the latest python, keeping miniconda version the same, because this is outside the scope of this PR

conda/meta.yaml
d3v-null: thanks, I've reintroduced the version constraint with a comment.

conda/conda_build_config.yaml
d3v-null: Let's see what happens if we get rid of this line. 
https://github.com/i4Ds/Karabo-Pipeline/actions/runs/16771087010

d3v-null: it's because it's building in an older docker container `ghcr.io/i4ds/mambabuild-docker:latest`

just need to update this one: https://github.com/i4Ds/mambabuild-docker 

Also a separate issue is that this step is currently failing because there's no python available in the environment, 

```
VERSIONEER_VERSION=$(python -c 'import versioneer; print(versioneer.get_version())')
```

d3v-null: I've just updated conda-build.yaml to match karabo-feedstock, please check that this looks ok https://github.com/i4Ds/Karabo-Pipeline/pull/653/files#diff-2c0e39bc014e056be855742e3ff5cc61ce0e4be5b6e1f4399af608b6aa22d8df 

I have a good feeling about this one: https://github.com/i4Ds/Karabo-Pipeline/actions/runs/16773792136/job/47494952070

Lukas113: Changes overall LGTM :+1: 

however, CI still fails (anaconda not available in base-env)

Lukas113: Yes, that's because `anaconda` is coming from `anaconda-client`, not from `anaconda` which is confusing XD.

In addition, I think the newest version of conda doesn't build `tar.bz2`, but `.conda` files.

d3v-null: Thanks Lukas, it's all working now! https://github.com/i4Ds/Karabo-Pipeline/actions/runs/16791257663 

I'd recommend squashing these commits.

Lukas113 (approved):
LGTM

PR #697, opened 2026-09-11
Sdp bug correction

LongPointyEyes:
Unit conversion in sdp imaging but corrected.
Updated docs after full removal of RASCIL


karabo/test/test_imager_sdp.py
anawas, line 1: Important new test case!

anawas (approved):
This looks good to me. It makes the api consistent with the ska-sdp package.

